PR #9, opened 2021-01-21
Feature/visualizations/chart interface

rkouye:


wmai (approved):
Great job! I realize that indeed the ChartJS documentation has its limits. I had to take a look at [DefinitelyTyped/chart.js](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a069b39865132e501f20587af39ba99c1ca596e6/types/chart.js/index.d.ts) to get a better idea of the configuration possibilities.

PR #12, opened 2021-02-01
Catch errors within ReactImpl to avoid higher-level violent crashes

richterb:
After running into fan-blowing, infinite-looping errors within our interfaces, I quickly built some system to render an "error!" block instead of just letting the error bubble up, and still try to render it again the next time. I'm not sure if we should handle the error block itself within the integration (translations, styling...), but the catching itself should happen there I think, as I don't think it's trivial to catch it up properly within the app (unless ErrorBoundary does the trick, including stopping rendering loops, I'm not entirely familiar with it yet). It's still very crude but it seems to save me a lot of time whenever I get a crash 😅


richterb:
Can be achieved easily within the calling application, no need to do this at this level. This leaves the ability for the application to catch errors however they see fit instead.

PR #19, opened 2021-03-10
Visualizations/add rollup visualizer and terser

rkouye:
Add size preview and minify for package visualizations.

Chart.js is actually 95% of the bundle size. We can reduce this by importing only what we use. I will take a look in another PR.

We can also add a GIthub action to monitor size variation like done for the api-client package.


richterb (approved):
👍
The bundle size evolution in PRs would be a great addition to avoid accidents in the future indeed

PR #20, opened 2021-04-01
Features/visualizations datalabels

rkouye:
Add chartjs plugin for datalabels (https://github.com/chartjs/chartjs-plugin-datalabels).

Both ChartJs 3 and this plugin are still in beta. It doesn't work yet with Pie and Radar, I will investigate later.

Let's wait for an official release to define an interface and document.

Screenshot 2021-04-01 at 14 11 11

### DIFF
--- a/packages/visualizations/package.json
+++ b/packages/visualizations/package.json
@@ -41,6 +41,8 @@
         "postcss": "^8.1.10",
         "rollup": "^2.33.1",
         "rollup-plugin-svelte": "^6.1.0",
+        "rollup-plugin-terser": "^7.0.2",
+        "rollup-plugin-visualizer": "^4.2.0",
         "svelte": "^3.29.4",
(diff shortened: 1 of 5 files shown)


packages/visualizations/package.json
```diff
@@ -41,15 +41,16 @@
...
+        "rollup-plugin-visualizer": "^4.2.0",
         "svelte": "^3.29.4",
         "svelte-check": "^1.1.8",
         "svelte-preprocess": "^4.6.1",
         "tslib": "^2.0.3",
         "typescript": "^4.0.5"
     },
     "dependencies": {
-        "chart.js": "3.0.0-beta.6",
-        "rollup-plugin-terser": "^7.0.2",
```
richterb, line 52: Ouch, nice catch 😅

richterb (approved):
👍 Looks nice!

PR #32, opened 2021-09-15
Layouts implementation, default formatters and pie charts custom line drawing plugin

RafaelSzmarowski:
### The goal for this PR is to : 

- Set all layouts listed types : https://preview.uxpin.com/240de15ef21cf5424ac691cea8dd934f19ab5699#/pages/138965156/comment/no-panels?mode=i

- Set charts designs : https://www.figma.com/proto/tg3Ap49vDnOsts5K6Iv7er/Untitled?page-id=0%3A1&node-id=2%3A2&viewport=360%2C48%2C1&scaling=min-zoom

- Creates default formatters (long ticks, legend and datalabels)

- Creates custom plugin to draw a line between datalabels and pie chart

Nota Bene : Fixing Typescript warnings is in progress


packages/visualizations/src/components/Chart/Chart.svelte
```diff
@@ -52,13 +164,51 @@
...
+                        'rgb(255,255,255)'
+                    );
+                }
+            },
+            offset: defaultValue(dataLabels.offset, 4),
+            borderRadius: defaultValue(dataLabels.borderRadius, 3),
+            formatter: formatter
+                ? (value, context) => {
+                      const formattedDataFrame = handleLongDataLabels(dataFrame);
+                      return formatter(context.dataIndex, formattedDataFrame);
```
wmai, line 206: I probably missed something but I'm a little bit confuse regarding `dataFrame`:

In `DataLabelsConfiguration` the `formatter` is defined as such:
```typescript
formatter?: (index:number, context:{ dataFrame:DataFrame }) => string | string[];
```

according to this, shouldn't we pass the argument this way instead?

```typescript
return formatter(context.dataIndex, { dataFrame: formattedDataFrame });
```

But after a few tests, it seems that the type declaration should change into:

```typescript
formatter?: (index:number, dataFrame:DataFrame) => string | string[];
```

package.json
```diff
@@ -30,6 +30,7 @@
         "new-version": "lerna version --conventional-commits --no-push"
     },
     "devDependencies": {
-        "lerna": "^3.22.1"
+        "lerna": "^3.22.1",
+        "svelte-check": "^2.2.6"
```
wmai, line 34: Why do we using the "next" version of this plugin, not the last one? Are there any features we use that are missing in this one?

wmai (commented):
How beautiful are our new graphs! 😍

My comments don't cover all the little problems of this PR, there are a lot of TypeScript errors to work out. A good group session would be great! 

I started working on the TypeScript errors in the `pieDataLabelsPlugin.ts` file on my local env. I manage to get rid of most of the errors but at the cost of some not very elegant "hacks" but I have a hard time understanding the chartJS typing in general. A little presentation of what you understood about this pluggin would certainly help me a lot to move forward.

rkouye:
🚀

PR #33, opened 2021-09-20
Add linting in package visualizations

rkouye:
Add linting in package visualizations and create a CI job to check it.

We use 3 mechanisms:
- **Prettier** for code formating. Dev can install prettier in their IDE to never have to think about code formatting anymore.
- **EsLint** for code style and best practices (configured for typescript and svelte, based on Airbnb style guide). The integration with IDE work quite well, however, in vs code I had to open the package folder as root (maybe I am missing something in my config) and configure the eslint plugin to work on svelte file.
- Finally, **Svelte Check** is added to check the typescript in Svelte before compilation, and some accessibility stuff.


wmai (approved):
We needed it indeed! Thanks!

PR #35, opened 2021-10-05
Fix refresh bug on the pie charts lines custom plugin

RafaelSzmarowski:
### The goal for this PR is to :

- Correct a bug noticed on the studio while changing the pie charts layouts. The custom plugin drawing lines was still applied until next page refresh (chartJS re-render). Registering the plugin globally and applying it only conditionally seems to resolve the observed bug.

### DIFF
--- a/packages/visualizations/src/components/Chart/index.ts
+++ b/packages/visualizations/src/components/Chart/index.ts
@@ -3,9 +3,11 @@ import ChartDataLabels from 'chartjs-plugin-datalabels';
 import type { ChartOptions, DataFrame } from '../types';
 import ChartImpl from './Chart.svelte';
 import SvelteImpl from '../SvelteImpl';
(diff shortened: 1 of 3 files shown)


packages/visualizations/src/components/Chart/index.ts
```diff
@@ -3,9 +3,11 @@ import ChartDataLabels from 'chartjs-plugin-datalabels';
 import type { ChartOptions, DataFrame } from '../types';
 import ChartImpl from './Chart.svelte';
 import SvelteImpl from '../SvelteImpl';
+import pieDataLabelsPlugin from './pieDataLabelsPlugin';
 
 ChartJs.Chart.register(...ChartJs.registerables);
 ChartJs.Chart.register(ChartDataLabels);
+ChartJs.Chart.register(pieDataLabelsPlugin);
```
rkouye, line 10: Nitpick , **P**ieDataLabelsPlugin

PR #44, opened 2021-11-04
MarkdownText: Display all lists as tight lists

richterb:
We only want to support tight lists, despite Markdown supporting both tight lists and loose lists.

This PR introduces a small plugin that makes it so MarkdownIt always renders tight lists, even if the initial parsing created a loose one (which is indicated by list items containing non-hidden paragraphs).


rkouye:
It may have been a good idea to add some visuals tests in the storybook.

PR #48, opened 2021-11-19
Add Tippy.js and tooltip for KPI

RafaelSzmarowski:
The goal for this PR is to implement Tippy.js and add tooltips for KPI value.

As for now, SDK user has to import tippy.css file directly on his own. Perhaps, a better implementation can be done on our side.


packages/visualizations/src/components/KpiCard/KpiCard.svelte
```diff
@@ -23,7 +24,7 @@
                 {#if options.title}
                     <h3 class="kpi-card__title">{options.title}</h3>
                 {/if}
-                <div class="kpi-card__value">
+                <div use:kpiTooltip={formattedValue} class="kpi-card__value">
```
etienneburdet, line 27: I think you can replace the whole use with 

```suggestion
    let valueNode: HTMLElement;
    let formattedValue = '';

    $: {
        formattedValue = data.value !== undefined ? data.value.toLocaleString() : '';
        tippy(valueNode, { content: formattedValue });
    }

    …

     <div bind:this={valueNode} class="kpi-card__value">
     ```
     And drop the use directive alltogether.

etienneburdet: Although don't forget to declare `let` variables before using it in reactive statements, else it's not observed in the same way.

rkouye: Quick late comment in case I forget on monday 🙈 .

The use directive allow you to destroy the tooltip when the node is removed. It is a better pattern IMO. When implemented correctly, you can reuse the existing instance and just update the content when the later change. Which is better than recreating a tippy instance every time the value change.

It is also more reusable, FYI @RafaelSzmarowski  I would have just named it `use:tippy`.

etienneburdet: Ha, true enough, if we have to tear destroy the instance it's much in a `use` than in the component `onDestroy`. I just found it a bit contrived to enforce title and aria on markup we wrote ourselves…

We just have to dedupe that logic in the action then (that's nitpicking, it's 4 lines of code…)

etienneburdet (approved):

PR #49, opened 2021-11-19
Add esm bundle in @opendatasoft/visualizations

rkouye:
Add an ESM bundle in package viz. 

It will be used by bundler when importing the library. 

It should also result in faster build time in dev env. And hopefully, lead to package deduplication in the studio.


wmai (approved):
I don't know much about rollup config, but since we do have a UMN and ESM bundle now, LGTM! Thanks

PR #50, opened 2021-11-24
Add css bundle in visualization

rkouye:


wmai (approved):
👏

packages/visualizations/rollup.config.js
```diff
@@ -61,38 +64,44 @@ const esm = {
...
+        production &&
+            visualizer({
+                filename: `gen/stats-es.html`,
+                sourcemap: true,
+            }),
+    ],
     onwarn,
 };
 
 const umd = {
```
etienneburdet, line 84: More a personnal question, but why do we still need to export UMD and ESM here ?

PR #51, opened 2021-11-24
Render chart title and subtitle outside canvas

rkouye:


RafaelSzmarowski (approved):
LGTM ! Well done and thanks for the huge work on the storybook to split everything individually 👍  
I just wonder if leaving an alignment option for the title/subtitle (start, center, end) won't be useful in the future.

PR #52, opened 2021-12-06
Feature: add link attributes based on patterns

etienneburdet:
## Description
The goal of this PR is to add attributes on link based on pattern and without extending markdown to hold this information.

## Changes
Added markdown-it-link-attributes and used it with a simple https:// pattern

## Open discussion
Should be pretty handy for buttons to.

## Review checklist

- [ ] Description is complete
- [ ] 2 squad members as reviewers (1 if trivial)
- [ ] A PM as reviewer for a functional testing
- [ ] A security reviewer for a change with security impacts or a story with a security tag
- [ ] Tests coverage has improved: _for a bug fix, there is a test such that, should the bug fix be reverted, this test would fail_
- [ ] Code is production ready : *errors*, *logs*, *memory*, *CPU*
- [ ] Code is production ready : impact on production in a year? what happens if a customer uses this feature too enthusiastically? using scripts? does complexity or storage size depend on its use over time?
- [ ] Code is production ready : can this code leak stuff in case of a crash at any time? database, objects, files, directories, temporary stuff?
- [ ] If I use some kind of secret (password, API key, app secret...): I commit the value "FAKE_SECRET", store the testing key in 1password, and document how to generate the production one -- I implement and document a way to rotate the secret without downtime


packages/visualizations/package.json
```diff
@@ -75,11 +75,13 @@
         "typescript": "^4.0.5"
     },
     "dependencies": {
+        "@types/markdown-it-link-attributes": "^3.0.1",
```
wmai, line 78: Types should rather be installed in `devDependencies`.

packages/visualizations-react/test/MarkdownText.test.tsx
```diff
@@ -47,4 +47,18 @@ describe('MarkdownText', () => {
...
+        [relative](/explore)
+        `
+        render(<MarkdownText options={{}} data={{ value }} />);
+        
+        const full = screen.getByRole('link', { name: 'full' });
+        const relative = screen.getByRole('link', { name: 'relative' });
+
+        expect(full).toHaveAttribute('target', '_blank');
+        expect(relative).not.toHaveAttribute('target', '_blank');
+    })
```
wmai, line 63: There's some linter errors in this file (in `MarkdownText.stories.tsx` as well)

wmai (changes requested):
Well done! Except for my comments about linting and type definition installation, LGTM!

etienneburdet:
@wmai @rkouye should be all good 😊 Still have missing types for Luxon, but I'm not sure where we are on this?

PR #56, opened 2022-01-04
Add exports endpoint + it's test

fpassaniti:
Expose API V2 /exports/ endpoint to the client.

Usage example from test app :
```
const getGeojson = async(domainid, datasetid, search = "", refine = {}, limit = 100) => {
    const client = new ApiClient({ domain: domainid });
    let query = fromCatalog().dataset(datasetid).exports('geojson').limit(limit);
    let keys = Object.keys(refine);
    for (let i = 0; i < keys.length; i += 1) {
        query = query.refine(`${keys[i]}:"${refine[keys[i]]}"`);
    }
    if (search) {
        query = query.where(`"${search}"`)
    }
    return client.get(query)
        .then(res => res)
        .catch(err => {
            throw err;
        });
};
```


rkouye:
LGTM ! thanks @fpassaniti

PR #58, opened 2022-01-05
!feat: Update api v2 client

rkouye:
Breaking change to follow the release of the API from BETA.

- Remove aggregates endpoints
- Add query endpoints
- Add list helper and optional parameters in Query methods


packages/api-client/src/odsql/index.ts
```diff
@@ -114,16 +129,17 @@ export class Query {
...
         facets: () => new Query(`${source}/facets/`),
         datasets: () => new Query(`${source}/datasets/`),
+        query: () => new Query(`${source}/query/`),
+        exports: (format: string) => new Query(`${source}/exports/${format}/`),
         dataset: (datasetId: string) => ({
             itself: () => new Query(`${source}/datasets/${datasetId}/`),
-            aggregates: () => new Query(`${source}/datasets/${datasetId}/aggregates/`),
             facets: () => new Query(`${source}/datasets/${datasetId}/facets/`),
             records: () => new Query(`${source}/datasets/${datasetId}/records/`),
+            query: () => new Query(`${source}/datasets/${datasetId}/query/`),
```
RafaelSzmarowski, line 140: Note: in query we need to pass `&include_metas=True` to get hateoas links

packages/api-client/src/odsql/index.ts
```diff
@@ -114,16 +129,17 @@ export class Query {
 function root(source: string) {
     return () => ({
         itself: () => new Query(`${source}/`),
-        aggregates: () => new Query(`${source}/aggregates/`),
         facets: () => new Query(`${source}/facets/`),
         datasets: () => new Query(`${source}/datasets/`),
+        query: () => new Query(`${source}/query/`),
```
RafaelSzmarowski, line 134: Note: in query we need to pass `&include_metas=True` to get hateoas links

packages/api-client/src/odsql/index.ts
```diff
@@ -86,12 +95,18 @@ export class Query {
...
+        );
     }
 
     offset(offset: NumberOrUpdater): Query {
         if (typeof offset === 'number') return this.set('offset', offset.toString());
-        return this.update('offset', (current: string) => offset(Number(current)).toString());
+        return this.update(
+            'offset',
+            offset && ((current: string) => offset(Number(current)).toString())
+        );
```
RafaelSzmarowski, line 109: Note: According to the API team provisional documentation, limit and offset will tend to be removed

RafaelSzmarowski (approved):
LGTM ! Good job 😃  I just left a few notes to leave some attention points

packages/api-client/src/odsql/index.ts
rkouye: I discussed with the api teams and we it is better to wait for those feature to be publicly documented first. Hateoas link don't really matter here yet, and are available in records/datasets endpoints.

PR #61, opened 2022-01-19
Add option to format chart tooltip

rkouye:


RafaelSzmarowski (approved):
LGTM ! It works well, I just have a small question, if I'm not mistaking, this only enables changing one part of the tooltip (the label displaying the value). Wouldn't be it better to enable a full custom tooltip and directly add a similar callbacks function to title so every part of the tooltip could be custom ?

rkouye:
We will probably enable a full custom tooltip, but so far we only need control on the label. There is a lot of options we can expose https://www.chartjs.org/docs/latest/configuration/tooltip.html , I am thinking that it may be wiser to add them as we progress.

PR #68, opened 2022-03-11
Maps add aspect ratio for maps container and canvas, handle resize and dynamic max zoom

RafaelSzmarowski:
The goal for this PR is to add an aspect ratio for maps.

The component will receive an aspect ratio that will resize the map container. To resize the canvas itself we will add an observer on the container and call map.resize() and map.fitBounds().

The maxZoom mechanism changes also a little bit, as some GEOJson might be duplicated we first merge bounding boxes for features with the same key and then compute the max zoom.


packages/visualizations/src/components/Map/MapRender.svelte
```diff
@@ -50,6 +58,22 @@ TODO:
             ...start,
         });
 
+        // Set a resizeObserver to resize map on container size changes
+        resizer = new ResizeObserver(debounce(() => {
+            map.resize();
+            // Cancel saved max bounds to properly fitBounds
+            map.setMaxBounds(null);
+            if (bbox) {
```
richterb, line 66: I'm not sure in which situations we don't have a bbox yet, but reading the code naïvely, I suppose the two calls to setMaxBounds could be inside the `if` clause, since they are only useful if we do `fitBounds`?

Also, once this is done, I guess theses three lines could be in a function, because the exact same lines are used in two places. Something like

```js
function fitMapToBbox(newBbox) {
    // Cancel saved max bounds to properly fitBounds
    map.setMaxBounds(null);
    map.fitBounds(newBbox, {
        animate: false,
        padding: 10,
    });
    // Rest min zoom and movement
    map.setMaxBounds(map.getBounds());
}
```

packages/visualizations/src/components/Map/MapRender.svelte
```diff
@@ -124,12 +148,15 @@ TODO:
...
+
 </script>
 
-<div id="map" bind:this={container} />
+<div id="map" bind:this={container} style={cssVarStyles} />
 
 <style>
     #map {
-        height: 400px;
+        aspect-ratio: var(--aspect-ratio);
```
richterb, line 160: [According to caniuse](https://caniuse.com/?search=aspect-ratio), browser support for `aspect-ratio` is really fresh. We're not supposed to support older browsers, but considering how young it is, and that by default it will default to a height of 0 if I remember correctly, we could try to have something ugly but showing something at least.

Not knowing at first how to do that, I discovered https://www.stevefenton.co.uk/2021/04/check-css-support-in-css-using-supports-rules/. Ironically, it depends on something that isn't support by all browsers (`@supports`), but it's much much wider. We could fallback to a fixed height of `400px` for example if `aspect-ratio` isn't support maybe? It's still ugly though, but otherwise, apart from doing calculations in JS and losing the benefits of `aspect-ratio`...

packages/visualizations/src/components/Map/MapRender.svelte
RafaelSzmarowski: Good remark ! I've moved the three lines into one function. As for existence check on bbox, I noticed that on map initialization the ResizeObserver can be called and then bbox is undefined.

packages/visualizations/src/components/Map/MapRender.svelte
RafaelSzmarowski: Indeed ! it's safer, I added an `@supports` fallback, the only case that won't be handled is if the browser doesn't support `aspect-ratio` and `@supports` in the same time...

I just wonder if something like this won't be even much safer :

packages/visualizations/src/components/Map/utils.js
```diff
@@ -53,65 +52,94 @@ export const mapKeyToColor = (values, colorScale) => {
...
+                feature.geometry.coordinates.forEach(coordsPath => {
+                    bbox = computeBboxFromCoords(coordsPath, bbox);
+                });
+                mergedBboxesArray.push(bbox);
+            } else {
+                feature.geometry.coordinates.forEach(coordsPath => {
+                    bbox = computeBboxFromCoords(coordsPath, bbox);
+                });
+                let storedBbox = mergedBboxesArray[uniqueIds.indexOf(id)]
+                const mergedBbox = [
```
etienneburdet, line 102: Maybe this can be factored in a function? Looks similar to the one above.

packages/visualizations/src/components/Map/utils.js
```diff
@@ -53,65 +52,94 @@ export const mapKeyToColor = (values, colorScale) => {
...
+            Math.max(coords[0], current[2]),
+            Math.max(coords[1], current[3]),
+        ]
+    }, bbox);
+}
+
+// Because features come from tiled vector data, feature geometries may be split
+// or duplicated across tile boundaries. As a result, features may appear
+// multiple times in query results.
+// To prevent looping on duplicated features in computeMaxZoomFromGeoJsonFeatures we merge
```
etienneburdet, line 80: Can we use `new Set(features)` to dedupe? If there are no differences in features at all… or `array.filter` else maybe?

etienneburdet (commented):
I just have a few syntactic remarks, else it looks good 👌 
As we discussed, it feels like we're battling against mapbox a bit too much, but I really lack for a better solution, so… 🤷

packages/visualizations/src/components/Map/utils.js
RafaelSzmarowski: It's not really similar as here we only set the merged bbox value between two bboxes whereas the other is creating a bbox from coordinates.

packages/visualizations/src/components/Map/utils.js
RafaelSzmarowski: 👍  I changed the function to avoid looping through arrays, it should be better now !

packages/visualizations/src/components/Map/utils.js
```diff
@@ -60,47 +59,83 @@ export const mapKeyToColor = (values, colorScale) => {
...
-        const bounds = new LngLatBounds(VOID_BOUNDS);
-        // FIXME: supports only shapes for now
-        if (feature.geometry.type !== 'Polygon') {
-            return;
-        }
-        feature.geometry.coordinates.forEach((coordsPath) => bounds.extend(coordsPath));
-
+    let maxZoom = 0; // maxZoom lowest value possible
+    const FilteredBboxesBis = mergeBboxFromFeaturesWithSameKey(features);
+    Object.entries(FilteredBboxesBis).forEach(([, value]) => {
```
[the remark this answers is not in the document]
RafaelSzmarowski: 👍

packages/visualizations/src/components/Map/utils.js
```diff
@@ -60,47 +59,83 @@ export const mapKeyToColor = (values, colorScale) => {
...
-        feature.geometry.coordinates.forEach((coordsPath) => bounds.extend(coordsPath));
+        feature.geometry.coordinates.forEach((coordsPath) => {
+            bbox = computeBboxFromCoords(coordsPath, bbox);
+        });
     });
-    return bounds;
+    return bbox;
 };
 
 export const computeMaxZoomFromGeoJsonFeatures = (mapContainer, features) => {
```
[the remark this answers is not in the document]
RafaelSzmarowski: Just added it 👍

richterb (approved):
Thank you, this feels very robust!

PR #70, opened 2022-03-18
Create first storybook stories for map choropleth

RafaelSzmarowski:
The goal for this PR is to create a first story of map choropleth in the storybook.


richterb (commented):
👌 Nice foundation to build upon, thanks

PR #78, opened 2022-04-15
Add legend component for choropleth

RafaelSzmarowski:
The goal for this PR is to create a simple legend for choropleth maps with two different options : gradient or palette. For now we don't give the possibility to make custom labels for the legend or change interpolation mode but this features will probably be implemented soon.

No formatting is applied as well for the numbers displayed in the legend, this will probably be part of a next PR.


packages/visualizations/src/components/Map/utils.js
```diff
@@ -1,19 +1,40 @@
...
-    const colorMin = chroma(colorScale).darken(4).hex();
-    const colorMax = chroma(colorScale).brighten(4).hex();
-
-    const scale = chroma.scale([colorMin, colorMax]).domain([min, max]);
+    let colorMin;
+    let colorMax;
+    let scale;
+
+    if (colorsScale?.type === 'palette') {
+        const tresholdArray = [];
```
richterb, line 17: Spelled "threshold" (thanks google, I knew my brain was reading something weird but I wasn't able to spell the right letters)

packages/visualizations/src/components/Map/utils.js
```diff
@@ -1,19 +1,40 @@
 import chroma from 'chroma-js';
 import geoViewport from '@mapbox/geo-viewport';
 
-export const colorShapes = (geoJson, values, colorScale) => {
+export const colorShapes = (geoJson, values, colorsScale) => {
     // Key in the values is "x"
     // Key in the shapes is "key"
     // We add a color property in the JSON
+    // FIXME Do we always want to take y or accept other keys
```
richterb, line 8: That's a good question. I always assumed we can enforce a convention where we want the data to have `x` and `y` (and maybe `label` for labels etc). I think it's reasonable, since here, worst case, you could write 2 lines of code to rename the keys before calling our own code.

Or did you have in mind a future situation where we could use more than one value in the calculations?

packages/visualizations/src/components/Map/utils.js
```diff
@@ -1,19 +1,40 @@
...
+        });
+        scale = chroma.scale(colorsScale.colors).classes(tresholdArray);
+    } else if (colorsScale?.type === 'gradient') {
+        colorMin = chroma(colorsScale.colors.start).hex();
+        colorMax = chroma(colorsScale.colors.end).hex();
+        scale = chroma.scale([colorMin, colorMax]).domain([min, max]);
+    } else {
+        // Default is basic grey color
+        const uniqueColor = chroma('#CBD2DB').hex();
+        scale = chroma.scale([uniqueColor, uniqueColor]).domain([min, max]);
```
richterb, line 36: I'm a bit confused because I would have expected this code to just return the same color for min and max 😅 

I think it may be easier to read (and possibly type later) if we just used a default value higher in the code (e.g. in MapRender, or in Choropleth if we want different default for different viz types later), like "if colorsScale is not defined, then we use a gradient between light grey and dark grey".

packages/visualizations/src/components/utils/ColorsLegend.svelte
```diff
@@ -0,0 +1,117 @@
...
+    import type { DataBounds, ColorsScale } from '../types';
+    import { defaultCompactLegendNumberFormat } from './formatter';
+    // options to customize the component
+    export let dataBounds: DataBounds;
+    export let colorsScale: ColorsScale;
+    export let clientWidth: number;
+    export let title: string;
+</script>
+
+<div class="legend-colors {clientWidth <= 375 ? 'legend-colors--fluid' : 'legend-colors--fixed'}">
```
richterb, line 13: I think it should at least be named `parentWidth` or something like that, to make it clear that it is the width of something outside the component.

I'm wondering if it shouldn't rather be up to the component that is using a legend (here, MapRender) to decide whether it want a fixed or fluid legend, using a "variant" property for example. It depends if we consider that the it's up to the legend itself to decide, and that it's always going to be the same rule for every component that uses it.

richterb (commented):
I have only played with it in Storybook, but I tried edge stuff (including things that currently don't work in the platform 🤫) and it is very robust and clean!

I have a few questions and remarks on the code style itself. All the logic itself seems clean, great work!

packages/visualizations/src/components/types.ts
```diff
@@ -217,3 +217,23 @@ export interface KpiCardOptions {
...
+    min: number;
+    max: number;
+}
+
+export type ColorsScale = GradientScale | PaletteScale;
+
+type GradientScale = {
+    type: 'gradient';
+    colors: {
+        start: string;
```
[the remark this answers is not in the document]
RafaelSzmarowski: Indeed, good point we already have `export type Color = string | string[];` so let's use it !

packages/visualizations/src/components/Map/utils.js
RafaelSzmarowski: Good catch ! I changed it 👍

packages/visualizations/src/components/types.ts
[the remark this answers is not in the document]
RafaelSzmarowski: I changed our `type Color = string | string[]` to `type Color = string` and replace everywhere we used it by `Color | Colo[]` I guess it makes it easier to read

packages/visualizations/src/components/Map/utils.js
RafaelSzmarowski: It was just a reminder to check if this will be a convention in the final version, and i guess we can indeed enforce this for now, I have no future situations in mind that would require something else, so I'll remove this comment !

packages/visualizations/src/components/utils/ColorsLegend.svelte
RafaelSzmarowski: Thanks for your remark ! As the legend component doesn't use the clientWidth (which good point should have been called parentWidth) I applied your second suggestion and passed a variant prop directly from MapRender to the legend component 👍

richterb (approved):
LGTM! 👏

PR #80, opened 2022-05-09
Change odsql date formating interface

rkouye:
⚠️  Api client breaking change

Javascript Date objects are not adequate to represent Date without timezone. Hence, the odsql date formatter interface no longer accept a Date object. The new interface require the user to pass year, month and day separately. To stay consistent with the API, we support `year`, `year-month` and `year-month-day` format.


packages/api-client/src/odsql/index.ts
```diff
@@ -164,7 +164,8 @@ export const string = (value: string) => JSON.stringify(value);
 
 export const dateTime = (date: Date) => `date'${date.toISOString()}'`;
 
-export const date = (date: Date) => `date'${date.toISOString().split('T')[0]}'`;
+export const date = (year: number, month?: number, day?: number) =>
+    `date'${year}${month !== undefined ? `-${month}` : ''}${day !== undefined ? `-${day}` : ''}'`;
```
richterb, line 168: Apparently the API V2 considers valid dates to be similar to ISO, so `2021-01` and not `2021-1` (https://public.opendatasoft.com/api/v2/catalog/datasets/?where=modified%3Edate%272021-1-1%27). 

We'd have to leftpad months and days. I heard there's a good package in NPM for that 😈. But since we're not supporting IE, `padStart` will work for us. Something like `String(day).padStart(2, "0")`.

Also, i'm thinking of situation where you already have the date as an ISO string (which should be frequent IMO, as most UI components can provide you with that). Maybe we can also allow taking a string directly and use it as-is? (it won't be as "structured", but it will still solve the issue of phantom timezones).
Otherwise, it's a bit tricky currently, as you'll have to do something like `date(...myIsoDate.split('-').map(Number))`

rkouye: If you have a date as a ISO string, you can directly insert it in the where clause normally.

etienneburdet:
Just added another util for people who would already have an iso string. We had it in the studio, but it's a very valid case and would be quite a burden to split.string, convert to number and reformat string.

wmai (approved):

richterb (approved):

PR #82, opened 2022-05-20
Refactor the calculation of initial bbox

richterb:
Previously we relies on `querySourceFeatures` or `queryRenderedFeatures` to recompute a bbox at the technical level (`MapRender`) regardless of the initial source. Unfortunately, we can only access geometries as they were re-optimized for rendering by Maplibre, meaning they are re-cut, discarded if outside the viewport or too small for the current zoom etc. In other words, we can't use it to access the original source data, so it's really not usable to compute the ideal bbox that represent all the data.

This PR:
- Changes `MapRender` to take a `bbox` parameter, which is not optional. `MapRender` is tasked with the rendering logic only, and it needs that now.
- `Choropleth` computes the bbox from the original GeoJSON data, and passes it to `MapRender`.
- Adds a few TODO where I'm not entirely sure we want to keep the current code. We should discuss this at some point, probably wen we have more real-life tests under our belt.
- I'm not sure yet how we'll deal with Vector Tiles. The cheapest way would be to require the bbox to be  passed alongside the VTiles URL (e.g. the bbox of France for the France cities vector tiles), but that won't work if you're only intereted in parts of the vector tiles (e.g. only the cities around Paris...). The solution may require enforcing the choice of a bbox in the initial configuration, but it's also intertwined with the UI we'd like to provide in the Studio. To be discussed once we're opening the subject I guess.

### DIFF
--- a/packages/visualizations/src/components/Map/MapRender.svelte
+++ b/packages/visualizations/src/components/Map/MapRender.svelte
@@ -12,10 +12,7 @@ TODO:
     import maplibregl from 'maplibre-gl';
     import { onMount } from 'svelte';
     import { debounce } from 'lodash';
-    import turfBbox from '@turf/bbox';
-    import {
-        computeMaxZoomFromGeoJsonFeatures,
-    } from './utils';
+    import { computeMaxZoomFromGeoJsonFeatures } from './utils';
 
     // maplibre style (basemap)
     export let style;
@@ -23,6 +20,8 @@ TODO:
     export let source;
     // maplibre layer config
     export let layer;
+    // bounding box to start from, and restrict to it
+    export let bbox;
 
     // aspect ratio based on width, by default equal to 1
     export let aspectRatio = 1;
@@ -30,8 +29,7 @@ TODO:
 
     let container;
     let map;
-    // bounding box to start from, and restrict to it
-    let bbox;
+
     let mapReady = false;
     // Used to add a listener to resize map on container changes, canceled on destroy
     let resizer;
@@ -82,12 +80,11 @@ TODO:
 
     function initializeResizer() {
         // Set a resizeObserver to resize map on container size changes
+        // TODO: Do we really want to reset to the initial bbox each time?
         resizer = new ResizeObserver(
             debounce(() => {
                 map.resize();
-                if (bbox) {
-                    fitMapToBbox(bbox);
-                }
+                fitMapToBbox(bbox);
             }, 100)
         );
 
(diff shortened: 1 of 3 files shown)


packages/visualizations/src/components/Map/MapRender.svelte
```diff
@@ -82,12 +80,11 @@ TODO:
...
     function initializeResizer() {
         // Set a resizeObserver to resize map on container size changes
+        // TODO: Do we really want to reset to the initial bbox each time?
         resizer = new ResizeObserver(
             debounce(() => {
                 map.resize();
-                if (bbox) {
-                    fitMapToBbox(bbox);
-                }
+                fitMapToBbox(bbox);
```
RafaelSzmarowski, line 87: I think here we should keep some checks before calling fitMapToBbox as I have this error in the console when using it in Studio only on the first render, it looks like fitMapToBbox is called before having all data.

Capture d’écran 2022-05-20 à 17 10 43

If you also encounter this error, something like this seems to be secure but a check on bbox could be enough:  
```
if (mapReady && bbox) {
     fitMapToBbox(bbox);
}
```

packages/visualizations/src/components/Map/MapRender.svelte
```diff
@@ -82,12 +80,11 @@ TODO:
 
     function initializeResizer() {
         // Set a resizeObserver to resize map on container size changes
+        // TODO: Do we really want to reset to the initial bbox each time?
```
RafaelSzmarowski, line 83: I guess it's really up to us and open for discussion, my feeling is that this works well and I'm afraid that someone resizing his window can be lost on a map with a huge zoom but I'm not closed to remove that as well.

RafaelSzmarowski (changes requested):
Good improvement for the calculation of the bbox 👍  and indeed there are some points that should be discussed in the near future !
LGTM ! I only have this error on the first render that I guess should be handled somehow (If of course I'm not the only one to have it).

RafaelSzmarowski (approved):
🚀

packages/visualizations/src/components/Map/MapRender.svelte
richterb: I added this comment because after my last "findings", I know that in some situations, queryRenderedFeatures / querySourceFeatures will not return every shape, and typically, if there is a situation where we have both large and very small shapes, we'll may start initially at a zoom where we won't get the smaller shape. So, we may compute a maximum zoom that will ignore smaller shapes and will actually be too wide.
But:
- This will happen 0.001% of the time?
- Even if this happen, it would require a very very very large difference of scale to give us a maximum zoom that is really too far from the small shape
- In other situations, it's rather cool in terms of usability

So I'm tempted to keep it as well and see if there's really a real-life situation that occurs. If it happens I'm not sure what we can do, but I think it won't...

PR #85, opened 2022-06-02
Add labels rotation for palette colors legend

RafaelSzmarowski:
The goal for this pull request is to add an automatic label rotation when the available width is reached by the size of the text.
hanges: 
- Bind clientwidth to check the available space and a function to rotate
- Also

### DIFF
--- a/packages/visualizations/src/components/utils/ColorsLegend.svelte
+++ b/packages/visualizations/src/components/utils/ColorsLegend.svelte
@@ -20,7 +20,7 @@
     // eslint-disable-next-line
     let labelRotationTimer: NodeJS.Timeout;
     const handleLabelRotation = (): void => {
-        let availableWidthPerLabel: number = legendWidth / numberOfLabels - 3;
+        const availableWidthPerLabel: number = legendWidth / numberOfLabels - 3;
         colorBoxWidth = legendWidth / numberOfLabels - 2;
         numberOfLabels = colorsScale.type === 'palette' ? colorsScale.colors.length + 1 : 2;
         maxLabelsSize = labelsWidth.reduce((a, b) => Math.max(a, b));
--- a/packages/visualizations/src/components/utils/ColorsLegend.svelte
+++ b/packages/visualizations/src/components/utils/ColorsLegend.svelte
@@ -1,6 +1,7 @@
 <svelte:options immutable={true} />
 
 <script lang="ts">
+    import { onDestroy, beforeUpdate } from 'svelte';
     import type { DataBounds, ColorsScale, LegendVariant } from '../types';
     import { defaultCompactLegendNumberFormat } from './formatter';
     // options to customize the component
@@ -8,6 +9,38 @@
     export let colorsScale: ColorsScale;
     export let variant: LegendVariant;
     export let title: string;
+
+    // the part below is related to labels rotation
+    let legendWidth: number;
+    let colorBoxWidth: number;
+    const labelsWidth: number[] = [];
+    let maxLabelsSize: number;
+    let numberOfLabels: number;
+    let availableWidthPerLabel: number;
+    let displayVertical: boolean;
+    // eslint-disable-next-line
+    let labelRotationTimer: NodeJS.Timeout;
+    const handleLabelRotation = (): void => {
+        availableWidthPerLabel = legendWidth / numberOfLabels - 3;
+        colorBoxWidth = legendWidth / numberOfLabels - 2;
+        numberOfLabels = colorsScale.type === 'palette' ? colorsScale.colors.length + 1 : 2;
+        maxLabelsSize = labelsWidth.reduce((a, b) => Math.max(a, b));
+        if (availableWidthPerLabel < maxLabelsSize) {
+            displayVertical = true;
+        } else {
+            displayVertical = false;
+        }
+    };
+    beforeUpdate(() => {
+        if (colorsScale.type === 'palette' && labelsWidth.length !== 0) {
+            clearTimeout(labelRotationTimer);
+            labelRotationTimer = setTimeout(() => {
+                handleLabelRotation();
+            }, 100);
+        }
+    });
+    // Clearing the setTimeout on destroy
+    onDestroy(() => clearTimeout(labelRotationTimer));
 </script>
 
 <div class={`legend-colors legend-colors--${variant}`}>
@@ -27,26 +60,50 @@
         </div>
     {:else if colorsScale.type === 'palette'}
         <!-- Palette color boxes, row display, no labels only displaying palettes steps -->
-        <div class="legend-colors-container-palette">
+        <div class="legend-colors-container-palette" bind:clientWidth={legendWidth}>
             <div class="legend-colors-row-color-box-palette">
                 {#each colorsScale.colors as color}
                     <div class="legend-colors-color-box-palette" style="--box-color: {color}" />
                 {/each}
             </div>
-            <div class="legend-colors-row-values-palette">
+            <div
+                class="legend-colors-row-values-palette {displayVertical
+                    ? 'vertical-labels-container'
+                    : ''}"
+            >
                 {#each colorsScale.colors as _color, i}
                     {#if i === 0}
-                        <div>{defaultCompactLegendNumberFormat(dataBounds.min)}</div>
-                        <div>
+                        <div
+                            bind:clientWidth={labelsWidth[i]}
+                            class:vertical-label={displayVertical}
+                            style="--label-row-width: {colorBoxWidth}px"
+                        >
+                            {defaultCompactLegendNumberFormat(dataBounds.min)}
+                        </div>
+                        <div
+                            bind:clientWidth={labelsWidth[i]}
+                            class:vertical-label={displayVertical}
+                            style="--label-row-width: {colorBoxWidth}px"
+                        >
                             {defaultCompactLegendNumberFormat(
                                 dataBounds.min +
                                     (dataBounds.max - dataBounds.min) / colorsScale.colors.length
                             )}
                         </div>
                     {:else if i === colorsScale.colors.length - 1}
-                        <div>{defaultCompactLegendNumberFormat(dataBounds.max)}</div>
+                        <div
+                            bind:clientWidth={labelsWidth[i]}
+                            class:vertical-label={displayVertical}
+                            style="--label-row-width: {colorBoxWidth}px"
+                        >
+                            {defaultCompactLegendNumberFormat(dataBounds.max)}
+                        </div>
                     {:else}
-                        <div>
+                        <div
+                            bind:clientWidth={labelsWidth[i]}
+                            class:vertical-label={displayVertical}
+                            style="--label-row-width: {colorBoxWidth}px"
+                        >
                             {defaultCompactLegendNumberFormat(
                                 dataBounds.min +
                                     ((dataBounds.max - dataBounds.min) /
@@ -113,5 +170,19 @@
     }
     .legend-colors-color-box-palette:not(:last-child) {
         margin-right: 1px;
+    }
+    /* Handle labels rotation */
+    .vertical-labels-container {
+        min-height: fit-content;
+        padding: 3px;
+        margin-top: 13px;
+        margin-left: -6px;
+        margin-right: -4px;
+        margin-bottom: 3px;
+    }
+    .vertical-labels-container > .vertical-label {
+        transform: rotate(270deg);
+        width: var(--label-row-width);
+        margin: auto;
     }
 </style>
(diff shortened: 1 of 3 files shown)


packages/visualizations/src/components/utils/ColorsLegend.svelte
```diff
@@ -27,26 +61,50 @@
...
-        <div class="legend-colors-container-palette">
+        <div class="legend-colors-container-palette" bind:clientWidth={legendWidth}>
             <div class="legend-colors-row-color-box-palette">
                 {#each colorsScale.colors as color}
                     <div class="legend-colors-color-box-palette" style="--box-color: {color}" />
                 {/each}
             </div>
-            <div class="legend-colors-row-values-palette">
+            <div
+                class="legend-colors-row-values-palette {displayVertical
```
richterb, line 71: Can't it be `class:vertical-labels-container="displayVertical"` like you did below? (I'm a Svelte noob)

packages/visualizations/src/components/utils/ColorsLegend.svelte
```diff
@@ -1,13 +1,47 @@
...
+    const rotationDebounce = debounce(handleLabelRotation, 200);
+    onMount(() => handleLabelRotation());
+    // Makes rotationDebounce reactive to legendWith and colorScale changes
+    $: if (legendWidth) {
+        rotationDebounce();
+    }
+    $: if (colorsScale) {
+        rotationDebounce();
+    }
+    onDestroy(() => rotationDebounce.cancel);
```
richterb, line 44: I think it should be either `onDestroy(rotationDebounce.cancel);` or `onDestroy(() => rotationDebounce.cancel());`

I guess the first one is nicer, and it could be `onMount(handleLabelRotation);` as well

richterb (changes requested):
This looks very nice!

Two issues: 
- It didn't seem to work for me on Firefox? (can't see why yet)
- At first load, you get a very small time with the horizontal labels, and then the vertical. What actually happens here:
  - The first call to `handleLabelRotation`, done in the `onMount`, somehow didn't detect that it should be vertical, and doesn't do anything
  - One of the two reactive statement suddenly activates a bit later, runs `handleLabelRotation` again, and this time it works
I think at first `legendWidth` is not there yet for some reason at `onMount` time, and then it appears a few milliseconds later, triggering the reactive statement. I don't know if there's something we could do, but ideally there shouldn't be a small jump. I'd suggest that we could investigate what exactly happens here, to make sure we don't run `handleLabelRotation`  too early, but still as soon as possible. If that's not enough, maybe we can delay displaying the labels for real until we're sure of our alignment.

packages/visualizations/src/components/utils/ColorsLegend.svelte
RafaelSzmarowski: 👍

RafaelSzmarowski:
> * It didn't seem to work for me on Firefox? (can't see why yet)

Interesting, on my Firefox it seems to work. Maybe a different version of Firefox (I'm on release 100.0) or an extension that blocks the cliendWidth information ?

packages/visualizations/src/components/utils/ColorsLegend.svelte
```diff
@@ -1,13 +1,47 @@
...
+    const labelsWidth: number[] = [];
+    let maxLabelsSize: number;
+    let numberOfLabels: number;
+    let displayVertical: boolean;
+    const handleLabelRotation = (): void => {
+        if (colorsScale.type === 'palette' && labelsWidth.length !== 0) {
+            const availableWidthPerLabel: number = legendWidth / numberOfLabels - 3;
+            colorBoxWidth = legendWidth / numberOfLabels;
+            numberOfLabels = colorsScale.type === 'palette' ? colorsScale.colors.length + 1 : 2;
+            maxLabelsSize = labelsWidth.reduce((a, b) => Math.max(a, b));
```
etienneburdet, line 27:
```suggestion
            maxLabelsSize = Math.max(...labelsWidth);
```
Super minor 😅

packages/visualizations/src/components/utils/ColorsLegend.svelte
```diff
@@ -1,13 +1,47 @@
...
     export let title: string;
+
+    // the part below is related to labels rotation
+    let legendWidth: number;
+    let colorBoxWidth: number;
+    const labelsWidth: number[] = [];
+    let maxLabelsSize: number;
+    let numberOfLabels: number;
+    let displayVertical: boolean;
+    const handleLabelRotation = (): void => {
```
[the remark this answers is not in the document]
etienneburdet: Here is a working exemple: 
```suggestion
        const handleLabelRotation = (legendW: number, labelW: number[]): void => {
        if (colorsScale.type === 'palette' && labelsWidth.length !== 0 && legendWidth) {
            console.log();
            const availableWidthPerLabel: number = legendWidth / numberOfLabels - 3;
            colorBoxWidth = legendWidth / numberOfLabels;
            numberOfLabels = colorsScale.colors.length + 1;
            maxLabelsSize = Math.max(...labelsWidth);
            return availableWidthPerLabel < maxLabelsSize;
        }
    };

    $: displayVertical = handleLabelRotation(legendWidth, labelsWidth);
```

What I like better in this pattern is that you have what triggers reactivity as argument and what is updated during the cycle on the left-handside.

I don't think debouncing is problem in itself as long as it has `legendWidth` as parameter and truly debounces the resize and is not used defensively against Svelte's cycles because we are triggering reactivity on the wrong variables.

packages/visualizations/src/components/utils/ColorsLegend.svelte
```diff
@@ -27,26 +61,49 @@
         </div>
     {:else if colorsScale.type === 'palette'}
```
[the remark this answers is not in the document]
RafaelSzmarowski: Yes ! Indeed, I guess that we will leave MapRender generic for all the maps (choropleth and markers map) So the legend component will probably be also generic with different types of legend and then we'll have to split everything 👍

packages/visualizations/src/components/utils/ColorsLegend.svelte
```diff
@@ -1,13 +1,47 @@
...
+    let colorBoxWidth: number;
+    const labelsWidth: number[] = [];
+    let maxLabelsSize: number;
+    let numberOfLabels: number;
+    let displayVertical: boolean;
+    const handleLabelRotation = (): void => {
+        if (colorsScale.type === 'palette' && labelsWidth.length !== 0) {
+            const availableWidthPerLabel: number = legendWidth / numberOfLabels - 3;
+            colorBoxWidth = legendWidth / numberOfLabels;
+            numberOfLabels = colorsScale.type === 'palette' ? colorsScale.colors.length + 1 : 2;
```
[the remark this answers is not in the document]
RafaelSzmarowski: 👍

packages/visualizations/src/components/utils/ColorsLegend.svelte
RafaelSzmarowski: 👍

packages/visualizations/src/components/utils/ColorsLegend.svelte
```diff
@@ -114,4 +184,16 @@
...
+        margin: 6px -6px;
+    }
+    .vertical-label {
+        transform: rotate(270deg);
+        height: var(--label-row-width);
+        min-width: var(--label-row-width);
+        line-height: var(--label-row-width);
+        margin: auto;
+        text-align: right;
+    }
```
etienneburdet, line 198:
```suggestion
    .legend-colors-row-color-box-palette,
    .legend-colors-row-values-palette {
        …
        line-height: 1em; /* for easier alignement */
    }
    
    ...
    .label-container {
        flex: 1 1 auto;
        display: flex;
        justify-content: center;
        align-items: flex-start;
    }

    .label-container span {
        margin-left: calc(-50% - 1em);
    }

    .vertical-label {
        writing-mode: vertical-lr; /* ensure parent height */
        transform: rotate(180deg);
    }
```
So the idea here is to use writing mode vertical, so that that the height is computed at layout time (contrary to transform: rotate) and it keeps the width in the "natural" direction. Then we offset to have all labels on the segment separators.

etienneburdet (changes requested):
So after playing a bit, I think the alignment method is not pixel perfect enoughg as of now. In this batch of suggestions, I tried a method where we add an empty transparente cell to have the same number of cells on both rows, align all labels to the center of the cells and then offset them. It also removes the need to force the line-height/width of the labels from computation, which can prove fragile.

It seems much more consistent, but needs more testing probably!

RafaelSzmarowski:
Thanks for your review and your changes ideas @etienneburdet ! I've tested your implementation proposal and indeed it makes the CSS more robust. However now the width and the height changes depending on the display (which is by the way more logic) so I need also to add a listener on clientHeight and compare the max labels height or width depending of the display. I guess that it shouldn't be a major performance issue ! Do not hesitate to give me your feedbacks 😃 We are not in a rush with this feature so we should make it as strong as possible !

packages/visualizations/src/components/utils/ColorsLegend.svelte
```diff
@@ -1,13 +1,63 @@
...
+        colorsScl: ColorsScale
+    ): void => {
+        if (isFirstRotation) {
+            handleLabelRotation(legendW, labelW, labelH, colorsScl);
+        } else {
+            rotationDebounce(legendW, labelW, labelH, colorsScl);
+        }
+    };
+
+    $: handleRotationLifecycle(legendWidth, labelsWidth, labelsHeight, colorsScale);
```
etienneburdet, line 59:
```suggestion

    $: if (labelsWidth > 0) {
        rotationDebounce(legendWidth, labelsWidth, labelsHeight, colorsScale);
      }
```
The problem we want to avoid here is calling rotationDebounce on the empty array and then wait before we have values. Instead of exposing a reactive state at component level—which is what I'm targeting with all those request 🙏 —, we just don't trigger unnecessary checks and thus debounce.

etienneburdet (changes requested):
Nicely done! Indeed we need to monitor height too now, since it is the "dimension" in some cases. We can still factor palette and gradient into their own component, but that might also be for later.

Except for the little change for initialization that avoid a state variable at component level, it looks good to me! 🚀

etienneburdet (commented):
**mistake**

packages/visualizations/src/components/utils/ColorsLegend.svelte
RafaelSzmarowski: Indeed ! { leading: true } seems to do the job ! I've made the change and I let you test it again 👍 Thanks for the tip !

etienneburdet (approved):
All good for me 😎 👌 Nicely done, it's way trickier than it looks… !

richterb:
Screenshot 2022-06-27 at 16 13 39

Is the offset on the right intended? It looks a bit weird, and in general I'd have said it looks like some `-1` missing somewhere because it's exactly the dimensions of a potential 6th spot here.

richterb:
The debounce is never called after the initial run, because it is designed to be called if the reference of `labelsWidth` and `labelsHeight` change, but they never change, they're initialized at the start. So if the data changes, the labels are never checked again.

Looks like adding ` && dataBounds` in the if line 47 may be enough for that.

RafaelSzmarowski:
> > Screenshot 2022-06-27 at 16 13 39
> > Is the offset on the right intended? It looks a bit weird, and in general I'd have said it looks like some `-1` missing somewhere because it's exactly the dimensions of a potential 6th spot here.
> 
> Well indeed, it looks like there is somewhere 1px missing only for the last label, I'll investigate that

I changed the margin-left of labels to align more the first and the last to the size of the boxes. It's a bit tricky as the formatting can be not the same for different labels and it's hard to align all labels between the edges... 
With the new changes it's seems maybe better, the pristine numbers are well aligned but the formatted ones in horizontal display have symmetrical margins between them but can be a little offset compared to the boxes, do you think it's better this way ? : 

Capture d’écran 2022-06-27 à 17 26 26


Capture d’écran 2022-06-27 à 17 26 15

Capture d’écran 2022-06-27 à 17 25 44
Capture d’écran 2022-06-27 à 17 25 55

packages/visualizations/src/components/utils/ColorsLegend.svelte
```diff
@@ -103,15 +170,25 @@
...
         display: flex;
-        margin-bottom: 3px;
+        margin-bottom: 6px;
     }
     .legend-colors-color-box-palette:not(:last-child) {
         margin-right: 1px;
     }
+    .legend-colors-container-palette .label-container {
+        display: flex;
+        justify-content: space-between;
```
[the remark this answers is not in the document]
RafaelSzmarowski: Thanks indeed !

richterb (approved):
Took a lot of work, and a final battle with Firefox, but we have a very good solution here 👏

PR #91, opened 2022-06-29
Reset map zoom when interactivity is disabled

RafaelSzmarowski:
The goal for this PR is to reset zoom on maps where interactivity is disabled, in case user has change default zoom and disable interactivity it avoids a frozen bad display of the map.


richterb (approved):
Works perfectly, clean and is very consistent from a functional point of view. Perfect solution 👌

PR #92, opened 2022-07-08
feat: pass more generic values to tooltip render

etienneburdet:
This PR aims at passing more values to toolitps in order allow for more rendering options, without having to rematch data/shapes manually.

## Open discussion
We pass the shapeKey, but maybe we could pass the dataKey (the "x") too? So that it's possible to do a fetch with more fields for exemples ?

### DIFF
--- a/packages/visualizations/src/components/Map/Choropleth.svelte
+++ b/packages/visualizations/src/components/Map/Choropleth.svelte
@@ -96,21 +96,24 @@ shapes: {
     }
 
     $: computeSourceLayerAndBboxes(data.value, shapes, colorsScale);
+
+    const defaultFormat = ({ value, label, shapeKey }) =>
+        value ? `${label} &mdash; ${value}` : label;
+
     $: renderTooltip = (hoveredFeature) => {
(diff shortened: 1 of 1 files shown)


richterb:
Wouldn't the dataKey and x be exactly the same by definition since it's the mapping key for both sides? (but then we can argue that we could just name it "key" then so that it's explicit that it's the same)

etienneburdet:
> Wouldn't the dataKey and x be exactly the same by definition since it's the mapping key for both sides? (but then we can argue that we could just name it "key" then so that it's explicit that it's the same)
Yes it would, that's the point 🤦 
They aren't that useful yet though since you still have to match an array, we don't index with those keys…

RafaelSzmarowski:
Seems to work well 👍 And it makes it easier to create the tooltip without to much computing and looping !

RafaelSzmarowski (changes requested):
LGTM ! We should just fix the stories on storybook before merging.

RafaelSzmarowski (approved):

richterb (approved):

PR #99, opened 2022-07-26
Update Chart.JS

richterb:
Also fix a long-standing warning regarding a redundant `role="link"` on an `<a>` element (which semantically already has that role).


richterb:
I tried to update all TS packages, but I got in a rabbit-hole, where the API-Client tests did fail because of the use of `new.target` in `error.ts` (to avoid duplicating the constructor in each sub-class) which ended up in an error:
```
/Users/benoit/dev/ods-dataviz-sdk/packages/api-client/src/client/error.ts:24
        var proto = _newTarget.prototype;
                    ^

ReferenceError: _newTarget is not defined
```
At some point I got stuck with the fact that TSDX isn't updated anymore, and I suspect one of the dependencies that didn't keep us with TS updates. I stopped there, and we'll see when we remove or replace TSDX at some point.

etienneburdet:
Yes Romaric was thinking of removing it too. It has some serious limitions when you want to do custom things in Jest too.

RafaelSzmarowski (approved):
LGTM !

PR #100, opened 2022-08-02
Moving all the map-related code to TypeScript

richterb:


packages/visualizations/src/components/Map/utils.ts
```diff
@@ -180,5 +183,5 @@ export const getFixedTooltips = (shapeKeys, features, renderTooltip) => {
         return null;
     });
 
-    return popups;
+    return popups.filter(Boolean) as ChoroplethFixedTooltipDescription[];
```
etienneburdet, line 186: So apparently [Boolean doesn't have the right signature for TS](https://github.com/microsoft/TypeScript/issues/16069). I really like the [isPresent ](https://www.npmjs.com/package/ts-is-present) solution, although we could craft our own function easily too.

Whatever the solution, I think we should do the extra effort to `popus.filter<ChoroplethFixedTooltipDescription>(…)` rather than hard casting. But this bug is a bit of pain indeed…

etienneburdet (changes requested):
Really nice, it will help a ton 😍  I have a few comments, but overall it's working great! 

The `any` parts are having me wondering if we should have a `no-explicit-any` rule. On the platform, it's a pain, but on a fresh codebase like this one it would be a nice reminder IMO (although it's easy enough). 

I really like the comments on types (didn't know they show in the IDE!), we should enforce that more in our code I think.

As I said, we should not forget to import the new Colorscale types in the platform too.

packages/visualizations/src/components/Map/utils.ts
richterb: I spent a lot of time trying to do it the right way, but somehow didn't succeed. I'll try using `compact` from `lodash` which should do the same thing, and has types that should work here.

etienneburdet: If we don't want to use the npm package `ts-is-present`, the other solution seems to be using [type guard returns](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates):
```
function nonNullable<T>(value: T): value is NonNullable<T> {
  return value !== null && value !== undefined;
}

return popus.filter<ChoroplethPopups>(nonNullable)
```

packages/visualizations/src/components/Map/types.ts
```diff
@@ -0,0 +1,71 @@
...
+    label,
+    key,
+}: {
+    /** Numeric value of the shape */
+    value?: number;
+    /** Label of the shape */
+    label: string;
+    /** Value of the key used to match shapes and numeric data */
+    key?: string;
+}) => string;
```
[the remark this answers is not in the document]
richterb: My reasoning here is that the SDK is exposing an option of this type, and typing even the return value makes sure that if the user passes a function that returns an int or a JSX.Element, their compiler/IDE complains that the SDK expects a function that always returns a string.

packages/visualizations/src/components/Map/types.ts
```diff
@@ -0,0 +1,71 @@
+import type { Feature, FeatureCollection, Position } from 'geojson';
+import type { FillLayerSpecification, Popup } from 'maplibre-gl';
+import type { Color, ColorsScale } from '../types';
+
+export interface ChoroplethOptions {
+    shapes: ChoroplethShapeValue;
+    colorsScale?: ColorsScale;
```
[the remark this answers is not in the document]
richterb: As discussed, there is a draft PR on the platform repo for that https://github.com/opendatasoft/platform/pull/9571, which makes the code nicer and tighter indeed

packages/visualizations/src/components/Map/MapRender.svelte
```diff
@@ -1,73 +1,85 @@
...
     // maplibre source config
-    export let source;
+    export let source: SourceSpecification;
     // maplibre layer config
-    export let layer;
+    export let layer: MapLayer;
     // bounding box to start from, and restrict to it
-    export let bbox;
-
+    export let bbox: BBox;
```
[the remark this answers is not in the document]
richterb: I was a bit torn here initially, because `LngLatBoundsLike` comes from Maplibre, and I'd rather expose things that are generic. But overall I think I prefer to require a parameter in a specific non-ambiguous format instead of tolerating many and having to support and sometimes transform them, and the bbox one is the most straight-forward (`[number, number, number, number]`).

packages/visualizations/src/components/Map/types.ts
```diff
@@ -0,0 +1,71 @@
...
+    center: Position;
+    description: string;
+    popup: Popup;
+}
+
+export type MapRenderTooltipFunction = (f: Feature) => string;
+
+export type ChoroplethLayer = Omit<FillLayerSpecification, 'id' | 'source'>;
+
+export type MapLayer = ChoroplethLayer;
```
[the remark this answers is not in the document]
richterb: In the code right now, we're literally just adding our own `id` and `source` and expecting the result to be a `FillLayerSpecification`, so this fits exactly the code and the intent, and if Maplibre evolves its specification, it's going to break instantly, which is nice. Right now I'm expecting that we won't have a lot of supported types of layer, but it really depends on how `MapRender` evolves. My experiences with Maplibre in general encourage me to really strenghten the configuration when we can 😅

packages/visualizations/src/components/Map/utils.ts
```diff
@@ -2,24 +2,37 @@ import chroma from 'chroma-js';
...
-    let scale;
+    let colorMin: Color;
+    let colorMax: Color;
+    let scale: Scale;
 
     if (colorsScale?.type === 'palette') {
-        const thresholdArray = [];
-        colorsScale.colors.forEach((_color, i) => {
+        const thresholdArray: number[] = [];
+        colorsScale.colors.forEach((_color, i: number) => {
```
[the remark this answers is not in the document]
richterb: Absolutely, I'm not sure why I went overboard here but not even with `_color` 😁

packages/visualizations/src/components/Map/utils.ts
```diff
@@ -36,15 +49,15 @@ export const colorShapes = (geoJson, values, colorsScale, emptyValueColor) => {
         scale = chroma.scale([colorMin, colorMax]).domain([min, max]);
     }
 
-    const dataMapping = {};
+    const dataMapping: any = {};
```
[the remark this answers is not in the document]
richterb: I tried the reduce, but it ends up being a bit more verbose because the initial value `{}` also needs the same type etc.
But I completely missed that I could have an object where both the key and value can be typed, so I replaced the `any` with the inlined type you suggested!

packages/visualizations/src/components/Map/utils.ts
```diff
@@ -103,8 +99,8 @@ function computeBboxFromCoords(coordsPath, bbox) {
 
 // The features given by querySourceFeatures are cut based on a tile representation
 // but we need the bounding box of the features themselves, so we need to build them again
-function mergeBboxFromFeaturesWithSameKey(features) {
-    const mergedBboxes = {};
+function mergeBboxFromFeaturesWithSameKey(features: Feature[]) {
+    const mergedBboxes: any = {};
```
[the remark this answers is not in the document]
richterb: Same as above, I didn't end up using a reduce because it complicated the code IMO, but I still replaced the any with a real inline type.

packages/visualizations/src/components/Map/utils.ts
richterb: I used an inline function:
```ts
return popups.filter((item): item is NonNullable<ChoroplethFixedTooltipDescription> => Boolean(item)) as ChoroplethFixedTooltipDescription[];
```

Thank you for teaching us about `is` 😁

packages/visualizations/src/components/Map/MapRender.svelte
```diff
@@ -113,9 +126,10 @@ TODO:
         return () => resizer?.disconnect();
     }
 
-    function sourceLoadingCallback(e) {
+    function sourceLoadingCallback(e: MapSourceDataEvent) {
         // sourceDataType can be "visibility" or "metadata", in which case it's not about the data itself
         if (e.isSourceLoaded && e.sourceId === sourceId && !e.sourceDataType) {
+            // @ts-ignore // The type forces you to pass a filter parameter in the option, but it's not required by the real code
```
wmai, line 132: There's an open issue on this subject: https://github.com/maplibre/maplibre-gl-js/issues/1393

Could we add a comment pointing to this issue in order to remove the `@ts-ignore` once they fix it?

packages/visualizations/src/components/Map/utils.ts
```diff
@@ -2,23 +2,36 @@ import chroma from 'chroma-js';
...
+
+export const LIGHT_GREY: Color = '#CBD2DB';
+export const DARK_GREY: Color = '#515457';
+
+export const colorShapes = (
+    geoJson: FeatureCollection,
+    values: ChoroplethDataValue[],
+    colorsScale: ColorsScale,
+    emptyValueColor: Color
+) => {
```
wmai, line 22: Since we will (maybe?) enforce a stricter typing checking on all packages, I'm reviewing this PR using a TS configuration similar to the one on the platform. In this context, the compiler complains because of the rule [`explicit-module-boundary-types`](https://github.com/typescript-eslint/typescript-eslint/blob/v4.31.2/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md) (same thing line 143).

But that's a problem for an other day ^^ I'm completely ok to leave it as it is. But if you want to change it:

```TypeScript
{
    geoJson: FeatureCollection,
    bounds: {
        min: number,
        max: number
    }
}
```

wmai (commented):
Very nice! Thanks @richterb 🙏

To be honest, I didn't want to delay this PR so I did a somewhat superficial review here. Maps are still a bit confusing for me, but I intend to dive into it and really understand what's going on.

packages/visualizations/src/components/Map/MapRender.svelte
```diff
@@ -137,7 +151,8 @@ TODO:
         }
     }
 
-    function addTooltip(e) {
+    function addTooltip(e: MapMouseEvent) {
+        // @ts-ignore // Somehow `features` isn't part of the type, but exists in the object at runtime
```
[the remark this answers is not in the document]
richterb: Yes! This was the thing I was looking for 😁 
I had a look at events specialized for hovering or something like that and didn't find it, but this is the one that I needed!

packages/visualizations/src/components/Map/MapRender.svelte
richterb: Thanks, didn't see that one!

packages/visualizations/src/components/Map/utils.ts
richterb: I didn't even know about that rule in the platform 😳 

I fixed it, better take good habits from the start!

packages/visualizations/src/components/Map/MapRender.svelte
```diff
@@ -1,73 +1,85 @@
...
     // option to disable map interactions
-    export let interactive;
-
+    export let interactive: boolean;
     // options to display legend
-    export let legend;
-    export let colorsScale;
-    export let dataBounds;
-    let clientWidth;
-    $: legendVariant = clientWidth <= 375 ? 'fluid' : 'fixed';
```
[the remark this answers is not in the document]
richterb: Indeed I can do somerhing like
```
const LEGENDVARIANT_FLUID = 'fluid';
const LEGENDVARIANT_FIXED = 'fixed';

export type LegendVariant = typeof LEGENDVARIANT_FLUID | typeof LEGENDVARIANT_FIXED;
```

But I'm not sure of the best place where to put the constants. I'd like to put them into ColorLegend.svelte, but I'm not sure how imports will work, and also this would be a circular import quickly.

What's the trick you're talking about exactly? Maybe it's precisely the answer to my problems, but I don't think I know about it

packages/visualizations/src/components/Map/MapRender.svelte
```diff
@@ -151,7 +153,7 @@
         }
     }
 
-    function addTooltip(e: MapMouseEvent) {
+    function addTooltip(e: MapLayerMouseEvent) {
         // @ts-ignore // Somehow `features` isn't part of the type, but exists in the object at runtime
```
wmai, line 157: I may have missed something, and I apologise in advance if that's the case. But couldn't we delete `@ts-ignore` and do a `if (e.features)` check instead?

richterb: Absolutely, you're right. Even my comment was outdated from before the `MapLayerMouseEvent` change, there was still an error for a different reason, I just didn't even look at it 😬

Changed in 20727d3d1480fa6113464d7592f370e7d4f53d5a

wmai (approved):
LGTM 👏

packages/visualizations/src/components/Map/Choropleth.svelte
```diff
@@ -1,18 +1,31 @@
 <svelte:options immutable={true} />
 
-<script>
+<script lang="ts">
     import turfBbox from '@turf/bbox';
+    import type { SourceSpecification } from 'maplibre-gl';
+    // eslint-disable-next-line import/no-unresolved
```
wmai, line 6: Regarding this issue (type only imports), I seems that the version `2.25.2` of `eslint-plugin-import` fixed it. I've tested this version and no linter more error 👍.

Here's the PR: https://github.com/import-js/eslint-plugin-import/pull/2220

(same thing in [MapRender.svelte](https://github.com/opendatasoft/ods-dataviz-sdk/pull/100/files#diff-39ec5d7e71c54fa06eac98981d148eace0b3a0ea67e9a19ca65ca535b93a93a4) line 16)

PR #104, opened 2022-09-01
New Choropleth component with SVG rendering, much lighter and quicker to render, but with no interactions

etienneburdet:
## Summary
The goal for this PR is to add an SVG map component to the SDK for drawing simpler maps more efficiently.

Associated Shortcut ticket : [sc-35449](https://app.shortcut.com/opendatasoft/story/35449/as-a-developer-i-have-an-svg-map-component-in-the-sdk-in-order-to-draw-small-static-simple-maps-without-loading-the-full).

## Changes
I created two components, one to render SVG maps with d3 geo (takes care of fitting to size, rendering etc.) and one for the choropleth (takes care of matching data and coloring shapes).

I created two subfolders in `Map`, one for SVG, one for webGL.

The storybook tests for reactivity on wrapper div, color scale and dynamic data, but I had no ideas for unit/integration tests.

I decided not to include the rewind utility for now as it seems to be related too much to the ODS api.

## Open discussion:
I tried to keep the map component as simple as possible for now, but the possibilities are [HUGE.](https://observablehq.com/collection/@d3/d3-geo). In a near future, it would be possible to : 
* Add tooltips
* Add multiples layers, e.g. a outline of regions with coloring of departments.
* Style stroke/outline
* Zoom/pan
* Change projection, add graticules etc.


packages/visualizations-react/stories/Map/SvgChoropleth.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+                padding: '13px',
+                display: 'inline-block',
+                height,
+                width,
+            }}
+        >
+            <SvgChoropleth
+                options={optionsWithScale}
+                data={dataSets[dataSet]}
+                style={{ height: '100%', width: '100%' }} // necessary to remove the wrapper div
```
RafaelSzmarowski, line 58: It's hard to understand why you need to pass this here, is it always needed ? I think it's a bit tricky to know that you need to add this here, can't it be a mandatory option of SvgChoropleth ?

RafaelSzmarowski (changes requested):
Good job ! It looks to work properly and the code is easy to read 👍  I have few remarks and few comments more open for discussion but overall LGTM. Nice storybook args controls !

packages/visualizations/src/components/Map/Svg/Choropleth.svelte
```diff
@@ -0,0 +1,26 @@
...
+
+    import type { SvgChoroplethOptions, ChoroplethDataValue } from '../types';
+
+    export let data: { value: ChoroplethDataValue[] };
+    export let options: SvgChoroplethOptions;
+    $: ({ geoJson, colorsScale = DEFAULT_COLORSSCALE, emptyValueColor = LIGHT_GREY } = options);
+
+    $: coloredShapes =
+        geoJson?.geoJson && geoJson.type === 'geojson'
+            ? colorShapes(geoJson.geoJson, data.value, colorsScale, emptyValueColor)
```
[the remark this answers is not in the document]
etienneburdet: The `mapKeyToColor` is actually all we need here, since we re-extract the color from the geoJson later on (with something a bit cumbersome tbh). We should wait for #105 to be merged and then move to `mapKeyToColor`.

packages/visualizations-react/stories/Map/SvgChoropleth.stories.tsx
etienneburdet: You mean the `style` part? It's because story book adds an empty wrapper div for some reason if there is no style. But it shouldn't be needed in any real code.

RafaelSzmarowski: Ah ok, wasn't aware of that. I've just tested removing it and the result was disturbing 😃 Maybe you can just add to the comment that this is a storybook specificity to be more explicit ?

packages/visualizations/src/components/Map/Svg/Choropleth.svelte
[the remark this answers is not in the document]
RafaelSzmarowski: 👍

etienneburdet:
@RafaelSzmarowski @richterb I based this PR on `vector-tiles` for now. This had me made quite a few changes, on top of adding stories, so probably worth a second round review 🙏

packages/visualizations/package.json
```diff
@@ -94,12 +95,17 @@
         "chartjs-plugin-datalabels": "^2.0.0",
         "chartjs-plugin-stacked100": "^1.2.0",
         "chroma-js": "^2.1.2",
+        "d3-geo": "^3.0.1",
+        "geojson-bbox": "^0.0.0",
+        "geojson-to-svg": "^1.1.4",
+        "geojson2svg": "^1.3.3",
```
richterb, line 101: I think these 4 packages were used for POCs and are not used anymore?
- `geojson-bbox`
- `geojson-to-svg`
- `geojson2svg`
- `reproject-spherical-mercator`

richterb: I removed them and I'll see if something breaks down the road 🙈

packages/visualizations/src/components/Map/Svg/Choropleth.svelte
```diff
@@ -0,0 +1,42 @@
...
+        shapes?.geoJson && shapes?.type === 'geojson'
+            ? colorShapes({
+                  featureCollection: shapes.geoJson,
+                  colorMapping,
+              })
+            : EMPTY_FC;
+</script>
+
+<Map
+    projection={geoEqualEarth()}
```
richterb, line 35:
```suggestion
    projection={geoMercator()}
```

D3 proposes the same projection used in every map on the web (Mercator). When I try it, I get more similar shapes, whereas the `geoEqualEarth` seem skewed to me (especially from DROM-COM that seems really streched)


Mercator
Screenshot 2022-10-17 at 15 25 27

Equal
Screenshot 2022-10-17 at 15 25 54

etienneburdet: Yep, that was the default projection, but geoMercator is the one used by Maplibre, so it makes more sense !

packages/visualizations/src/components/types.ts
```diff
@@ -34,4 +34,4 @@ export type PaletteScale = {
     colors: Color[];
 };
 
-export type ColorScales = GradientScale | PaletteScale;
+export type ColorScale = GradientScale | PaletteScale;
```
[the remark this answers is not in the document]
etienneburdet: I had a case where I merged both `ColorsScale` and `ColorScales` from diffrent branches, so I was like "screw it, I make my own!" 😆 And then we have to fix it everywhere 🤷

packages/visualizations/src/components/Map/Svg/Choropleth.svelte
```diff
@@ -0,0 +1,42 @@
...
+        EMPTY_FC,
+        getDataBounds,
+        colorShapes,
+    } from '../utils';
+    import Map from './Map.svelte';
+
+    import type { GeoJsonChoroplethOptions, ChoroplethDataValue } from '../types';
+
+    export let data: { value: ChoroplethDataValue[] };
+    export let options: GeoJsonChoroplethOptions;
```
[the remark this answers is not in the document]
etienneburdet: It needs a fixed height/width somewhere, I assumed it could be the wrapper, but yes if some uses % everywhere, or flex, it will get weird.

richterb (approved):
👏 This is going to make great DROM-COM composite maps

PR #105, opened 2022-09-05
Support vector tiles source in Choropleth's SDK component

RafaelSzmarowski:
**Do not merge until all the needs in @wmai implementation are covered**

The goal for this PR is to integrate vector tiles choropleths.

**Changes :** 

- New Choropleth possible source : vector tiles url
- Change label props to labelFormatter to be more explicit
- Add a labelMatcher props that can be of two types: select a target shape property that will be the label or pass an object of shapes key and labels to map on. It allows users to pass more custom labels and is specially usefull with vector tiles as in some cases it's hard to change vector tiles properties (static external vector tiles files for example).
- New props common both for GeoJson and Vector Tiles: fixedBbox (force a bbox and override the automatic computed one for geoJson), filter (decides which features render)
- Allow displaying shapes with an empty data object (before it was crashing)
- Add a debounce on the tooltip for a slight improvement
- Split typescript types, one file for each component, keep a higher level types.Ts file only for types that can be common to multiple components
- Strengthen types through enums and checking if switch is exhaustive

**To be discussed :** 

- It doesn't seem easy to compute a proper Bounding Box and MaxZoom level from the vector tiles properties. 
  - reasons: we can only get properties from MapLibre after the source callback, a small delay will appear on loading the map, the geometry prop has an underscore in MapLibre so it seems not stable, it's hard to know the type of features received, in our tests there are Polygons and MultiPolygons and it seems that Turf only works with geojson, doing all the computing manually would have implied to join all tiles with same id together but the fact that there are also MultiPolygons with same ids could cause difficulties. We could use a library to convert vector tiles to geoJson (vector-tile-js or vt-geojson) if there is a strong necessity of having an automatic bbox and maxzoom but it seems maybe to heavy for what we want to achieve.
 
  Should we force users to always pass a fixedBbox and maxZoom props for vector tiles maps ? For now fixedBbox is optionnal (default is the biggest bbox) and there's no maxZoom for vector tiles.

- Should we split Choropleth into two separate components (for example: ChoroplethGeoJson and ChoroplethVectorTiles). It doesn"t seem like something obvious for now and the switch is not so heavy inside one component but it's just an opinion.

- For now, we use map.setFilter to filter the vector tiles we want to display and the data values object to sync the legend scope. It seems to be better than always render all features and then color not desired features with the matching background color and filter adding a tooltip to hide them.

- Is there a way to add a feature property for the tooltip label directly on the vector tiles ? Until now we had to use a property that is coming from the feature. I added a new labelMatcher props to make custom labels.

- The filter props only accepts an array of values which seems sufficient for our need. Should we make this props already more complex by adding a targeted layer, targeted feature property on which filter (ex: key, value) and the corresponding array of values ?


packages/visualizations/src/components/Map/Choropleth.svelte
```diff
@@ -2,58 +2,60 @@
...
-
-    const defaultEmptyValueColor = '#cccccc';
+    let shapes: ChoroplethShapeValues;
+    let colorsScale: ColorScales;
 
     let aspectRatio: number;
     let renderTooltip: MapRenderTooltipFunction;
-    let bbox: BBox;
+    let bbox: BBox | undefined;
+    let fixedBbox: BBox | undefined;
```
richterb, line 33: Having both `bbox` and `fixedBbox` seems too complicated IMO.
I'm not sure what `bbox` is used for now.

Can't we just have `bbox`, and if it's set, use it, else compute one if we can (= geojson)?

packages/visualizations/src/components/Map/MapRender.svelte
```diff
@@ -129,23 +133,34 @@
...
+                }
+            );
             if (renderedFeatures.length) {
-                // Restrict zoom max
+                // Restrict zoom max only works for geojson for now
                 // TODO: We may not catch the smaller shapes if Maplibre discarded them for rendering reasons, so it's a bit risky. Is it worth it?
                 // A low-cost approach could be to restrict the zoom scale to an arbitrary value (e.g. only 4 from the max zoom)... or not restrict at all.
-                const maxZoom = computeMaxZoomFromGeoJsonFeatures(container, renderedFeatures);
-                map.setMaxZoom(maxZoom);
+                if (e.source.type === 'geojson') {
```
richterb, line 150: I'd be tempted to also support activeShapes only for GeoJSON sources. I feel like it's going to be very wonky. Then the entire `sourceLoadingCallback` (and therefore the fact that we listen to `sourcedata` in the first place) probably only makes sense for GeoJSON.

packages/visualizations/src/components/Map/types.ts
```diff
@@ -1,16 +1,22 @@
...
     interactive?: boolean;
-    emptyValueColor: Color;
-    tooltip: { label: ChoroplethTooltipFormatter };
+    emptyValueColor?: Color;
+    tooltip?: {
+        labelFormatter?: ChoroplethTooltipFormatter;
+        labelMatcher?: ChoroplethTooltipMatchers;
+    };
+    bbox?: BBox | undefined;
+    filter?: string[] | number[] | undefined;
```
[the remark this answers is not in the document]
wmai: Yes this interface looks good to me 👍

> I suggest we implement the bare minimum we need for now

I agree. And it won't be difficult to allow array of values in the future if needed.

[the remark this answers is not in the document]
etienneburdet: Agreed too, but maybe `filter?: { [key: string]: string }`. Or is it harder to manipulate afterward ?

packages/visualizations/src/components/Map/Choropleth.svelte
```diff
@@ -63,44 +64,74 @@
...
+            layer = {
+                type: 'fill',
+                'source-layer': newShapes.layer,
+                layout: {},
+                paint: {
+                    'fill-color': fillColor,
+                    'fill-outline-color': DEFAULT_COLORS.ShapeOutline,
+                    'fill-opacity': [
+                        'case',
+                        ['boolean', ['feature-state', 'hover'], false],
```
richterb, line 127: Why this rule, and if we want something like this, shouldn't it exist for GeoJSON as well?

Also it doesn't seem to have any effect in the Storybook

packages/visualizations/src/components/Map/MapRender.svelte
richterb: Since we're using activeShapes for test, I just added a warning in the comments, and we may even want to not document that at all in the future. But it's useful for us to keep it for now indeed.

packages/visualizations/src/components/Map/Choropleth.svelte
wmai: Oh I think that was some experimentations of mine that wasn't supposed to be committed. My bad.,

richterb: I'll set 0.8 like for GeoJSON and mutualize the style a bit in the code

packages/visualizations/src/components/Map/types.ts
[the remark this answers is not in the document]
richterb: I finally kept support for arrays, because otherwise I'd have to rewrite all tests to use départements instead of régions 😬 It's 10 characters more in the code so I think it's alright

packages/visualizations/src/components/Map/utils.ts
```diff
@@ -119,11 +101,9 @@ function mergeBboxFromFeaturesWithSameKey(features: Feature[]) {
             feature.geometry.coordinates.forEach((coordsPath) => {
```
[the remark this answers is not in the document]
richterb: It seems properly introspected in my VS Code somehow, and I don't have that warning. Maybe you have a different version of TS, with all the refactoring going on? But it would be easy to explicitly type this.

Screenshot 2022-09-20 at 11 56 52

richterb:
> > Should we force users to always pass a fixedBbox and maxZoom props for vector tiles maps ? For now fixedBbox is optionnal (default is the biggest bbox) and there's no maxZoom for vector tiles.
> 
> I don't think we can find a safe way to compute these for VTiles, so something like that IMO yes. The potential issue is that for a single component (`Choropleth`), the semantics of the options change for some properties like `bbox` and `maxZoom`: for GeoJSON, we'll compute them, for VTiles, we can't (and so bbox will default to the world I guess).

The more I think about it, the more I think we could just ask for a `bbox` and a `maxZoom`, defaulting to the world and no max zoom if unset. And on the Studio's side, compute what we can (bbox and maxzoom for geojson) using the same code. This makes the interface much more simple and consistent, and in real life, the need for his kind of magic generic things doesn't seem obvious. Probably not part of this PR though... But before release IMO.

richterb (approved):
Looks good for a first round!

To do for a later time:
- discuss the need for labelMatcher: I think just having `label: 'name of the field to use"` would be much simpler in the configuration, if we don't need all the logic anymore
- compare performance of explicit color-value matching vs letting Maplibre compute the gradient itself: if it's much quicker (e.g. for 36K cities) and does exactly the color gradient algo we need, then we could use it.
- move magic bbox and maxzoom in Studio?

richterb:
⚠️ Let's not merge until all the TSDX and packaging rework has been done, so that we do a last minor release with all the heavy lifting, and then a major with just the map changes.

etienneburdet:
@richterb @RafaelSzmarowski I just merged main, with a bit of impact (like an S in ColorScales that had two versions…). Probably worth a round of rereview just to be sure since change were a bit heavy 👌

PR #106, opened 2022-09-06
Chore/remoe-tsdx-api-client

etienneburdet:
Goal of this PR to remove TSDX for the api-client package (only).

This PR is only for the API client, in a plan to have further PR vizualisation-react and then another one to factor what we can at top level.

## Changes
Added rollup and all necessary dependencies for jest and rollup.
Removed TSDX


wmai (approved):
That's great! Thank you very much for this! 🙏

Little by little we will manage to get rid of TSDX. 💪

RafaelSzmarowski (approved):
LGTM ! Thanks for this improvement !

wmai:
🚀

wmai:
This PR was merge in the wrong branch, here's the effective merge: 505241a

packages/api-client/rollup.config.js
```diff
@@ -0,0 +1,107 @@
...
+            visualizer({
+                filename: 'gen/stats-umd.html',
+                sourcemap: true,
+            }),
+    ],
+    onwarn,
+});
+
+// Just compile the ESM version during development
+const bundles = production ? [esm, umd] : [esm];
```
richterb, line 105: Previously we also did build a CJS version, is there a conscious to not do that anymore? As far as I understood it can still be useful in some situations, and doesn't cost much to build for production.

packages/api-client/rollup.config.js
```diff
@@ -0,0 +1,107 @@
...
+                sourcemap: true,
+            }),
+    ],
+    onwarn,
+});
+
+const umd = defineConfig({
+    input: 'src/index.ts',
+    output: {
+        dir: 'dist',
```
richterb, line 77: In `package.json` we have
```
"unpkg": "umd/opendatasoft.apiclient.umd.js",
```
Apparently unpkg requires the build to be in an `umd` folder, hence the previous location: https://unpkg.com/#workflow
I guess we should keep the UMD build in a separate `umd` folder still?

packages/api-client/rollup.config.js
```diff
@@ -0,0 +1,107 @@
...
+    warn(warning);
+}
+
+const esm = defineConfig({
+    input: 'src/index.ts',
+    // Externalize all dependencies
+    external: (id) => Object.keys(pkg.dependencies).includes(id),
+    output: {
+        dir: 'dist',
+        entryFileNames: '[name].es.js',
```
richterb, line 58: It's still expected to be named `.esm.js` in `package.json`:
```
"module": "dist/api-client.esm.js",
```
That being said, I can't find any documentation on a `module` property for package.json files, so I suppose we can just remove that line instead 😄

packages/api-client/rollup.config.js
```diff
@@ -0,0 +1,107 @@
...
+import pkg from './package.json';
+
+const production = !process.env.ROLLUP_WATCH;
+
+function basePlugins() {
+    return [
+        typescript({
+            sourceMap: true,
+            declaration: true,
+            declarationDir: 'dist',
```
richterb, line 21: The `typings` property in `package.json` still targets `dist/index.d.ts` and should be updated (now `dist/src/index.d.ts`)

richterb (commented):
A bit late to the party 😅 

I saw a few issues with `package.json`,and one question about CJS.

Using `npm run size` (with an updated version because the current one doesn't work, I'll make a PR for it), I saw the new files are heavier than the previous:

Before:
Screenshot 2022-09-14 at 14 45 51

After:
Screenshot 2022-09-14 at 14 49 30

In theory there is no reason for this as far as I can tell, unless I missed other changes. It should be investigated IMO.

(obviously future changes should be in a new PR, it was just easier to point out the changes here)

packages/api-client/rollup.config.js
richterb: After looking at CJS with noob eyes, it looks like UMD covers all the cases that CJS does. Let's not do it then, and if needed we'll see how and why.

PR #108, opened 2022-09-14
Chore: Remove tsdx from visualization-react

etienneburdet:
## Description
The goal of this PR is to remove TSDX from `visualization-react` in a similar fashion than #106

## Changes
Added rollup, eslint and jest config, as close as possible to the ones we use in platform and other packages. I had to bump up some package versions as well for everytring to work together. TS is purposefully 4.6 where an update would lead to 4.7+, because 4.7 has [a few breaking changes](https://github.com/microsoft/TypeScript/wiki/Breaking-Changes#typescript-47) our dependencies don't support yet (mainly storybook with the `{...args}`).

I had to change some stories a bit as some utility functions we used made typing cumbersome or unsafe (such as `storyWithArgs`), where storybook has utility types that do the job (such as `ComponentStory<typeof T>`). Some of the stories also had wrong typing in their args, which is now corrected.

I tried to change as few code as possible, but I think some of our stories are still doing complicated things that would better be served by Storybook's controls and parameters. Kpi/Studio is a prime exemple, the "withData" utititilies and format function options could just be controls.

## Open discussion
The rollup config could be improved (like cjs), but I'm of the opinion to leave that for a further PR—this one works and already has quite a big diff. But there are many areas to improve, such as type exporting, CJS/UMD etc., analyze bundle size etc.


etienneburdet:
Closing due to rebase/git hierchy problem in favor of #113 (which his just all commits squashed and cherry-picked onto main).

PR #110, opened 2022-09-20
API Client fixes after TSDX removal

richterb:
- changed paths in `package.json`
- moved UMD build into `umd` folder as before because [unpkg seems to expect it](https://unpkg.com/#workflow)
- changed the name of the global object from `opendatasoft.visualization` to `opendatasoft.apiclient` for ease of use

### DIFF
--- a/packages/api-client/package.json
+++ b/packages/api-client/package.json
@@ -13,10 +13,9 @@
         "JavaScript",
         "TypeScript"
     ],
-    "main": "dist/index.js",
-    "typings": "dist/index.d.ts",
-    "module": "dist/api-client.esm.js",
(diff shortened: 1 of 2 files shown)


etienneburdet (approved):
LGTM! I indeed see no use for CJS in a fresh package like this, there is no chance it's a deep dep somewhere in a legacy app.

As for the size, I see multiple candidates potential candidates: different rollup ts plugin (with different options), possibly a few extra config steps, even the new TS version… 

I'm not that concerned about the size of the package itself, it shouldn't impact the end user if treeshaked well (that seems to be the case on the demo app).

PR #113, opened 2022-09-22
Chore: remove tsdx and replace with rollup/jest for visualizations-react

etienneburdet:
## Description
The goal of this PR is to remove TSDX from `visualization-react` in a similar fashion than #106

## Changes
Added rollup, eslint and jest config, as close as possible to the ones we use in platform and other packages. I had to bump up some package versions as well for everytring to work together. TS is purposefully 4.6 where an update would lead to 4.7+, because 4.7 has [a few breaking changes](https://github.com/microsoft/TypeScript/wiki/Breaking-Changes#typescript-47) our dependencies don't support yet (mainly storybook with the `{...args}`).

I had to change some stories a bit as some utility functions we used made typing cumbersome or unsafe (such as `storyWithArgs`), where storybook has utility types that do the job (such as `ComponentStory<typeof T>`). Some of the stories also had wrong typing in their args, which is now corrected.

I tried to change as few code as possible, but I think some of our stories are still doing complicated things that would better be served by Storybook's controls and parameters. Kpi/Studio is a prime exemple, the "withData" utititilies and format function options could just be controls.

## Open discussion
The rollup config could be improved (like cjs), but I'm of the opinion to leave that for a further PR—this one works and already has quite a big diff. But there are many areas to improve, such as type exporting, CJS/UMD etc., analyze bundle size etc.


packages/api-client/rollup.config.js
```diff
@@ -75,10 +75,10 @@ const umd = defineConfig({
     input: 'src/index.ts',
     output: {
         dir: 'dist',
-        entryFileNames: 'umd/[name].umd.js',
+        entryFileNames: '[name].umd.js',
         format: 'umd',
         sourcemap: true,
-        name: 'opendatasoft.apiclient',
+        name: 'opendatasoft.visualizations',
```
richterb, line 81: I think somehow your PR reverted another one of my PR: https://github.com/opendatasoft/ods-dataviz-sdk/pull/110

Maybe something with all the rebase funk? 😅 But if it's unvoluntary, I think it's probably easier to just re-commit the 2 lines that changed, and not jump back into Git problems...

etienneburdet: No I might have really CC there. I'll double check the output directories too.

packages/visualizations-react/package.json
```diff
@@ -47,31 +53,60 @@
     },
     "dependencies": {
         "@opendatasoft/visualizations": "^0.9.0",
+        "@testing-library/jest-dom": "^5.11.9",
+        "@testing-library/react": "^11.2.5",
+        "@types/jest": "^29.0.0",
```
richterb, line 58: I would have expected these 3 libs should be in `devDependencies`, not in `dependencies`, as they should absolutely not be packaged in regular builds. Is there a specific reason?

packages/visualizations-react/package.json
```diff
@@ -47,31 +53,60 @@
...
+        "@rollup/plugin-babel": "^5.3.0",
+        "@rollup/plugin-commonjs": "^19.0.0",
+        "@rollup/plugin-json": "^4.1.0",
+        "@rollup/plugin-node-resolve": "^10.0.0",
+        "@rollup/plugin-replace": "^3.0.1",
+        "@rollup/plugin-typescript": "^6.1.0",
+        "@size-limit/preset-small-lib": "^8.1.0",
         "@storybook/addon-essentials": "^6.4.22",
-        "@storybook/addon-info": "^5.3.21",
+        "@storybook/addon-info": "^4.1.18",
```
richterb, line 74: Why the downgrade?

But moreover, it seems `addon-info` has been deprecated and replaced with `addon-docs`: https://github.com/storybookjs/deprecated-addons/tree/master/addons/info . I'm not sure why it is used for and if it's a drop-in replacement though.

packages/visualizations-react/package.json
```diff
@@ -47,31 +53,60 @@
...
         "@storybook/addon-links": "^6.4.22",
         "@storybook/addons": "^6.4.22",
-        "@storybook/react": "^6.4.22",
-        "@testing-library/jest-dom": "^5.11.9",
-        "@testing-library/react": "^11.2.5",
-        "@types/jest": "^26.0.20",
+        "@storybook/react": "^6.1.21",
```
richterb, line 77: Why the downgrade? Feels weird to have a different version that one from storybook

packages/visualizations-react/package.json
```diff
@@ -47,31 +53,60 @@
...
+        "jest-environment-jsdom": "^29.0.2",
+        "jest-fetch-mock": "^3.0.3",
+        "npm-run-all": "^4.1.5",
+        "postcss": "^8.4.16",
         "react": "^16.13.0",
         "react-dom": "^16.13.0",
         "react-is": "^16.13.0",
-        "size-limit": "^4.9.1",
-        "tsdx": "^0.14.1",
+        "rollup": "^2.33.1",
```
richterb, line 104: Was it the version used by TSDX?

It is 2 years old now, I agree we should stick to the TSDX one in this PR, but in the near future, bumping rollup across our 3 packages could be nice. I suppose there were a loooot of bugfixes and improvements. (I'm not sure how many breaking changes though 😅)

packages/visualizations-react/rollup.config.js
```diff
@@ -0,0 +1,107 @@
...
+});
+
+const umd = defineConfig({
+    input: 'src/index.tsx',
+    output: {
+        dir: 'dist',
+        entryFileNames: '[name].umd.js',
+        format: 'umd',
+        sourcemap: true,
+        name: 'opendatasoft.visualizations',
```
richterb, line 81: I'm actually wondering if an UMD build makes sense for a React library. For either way it should expose a different name to avoid conflicts here, such as `opendatasoft.visualizationsReact` (low-effort name 😅)

richterb (commented):
Huge work on the Stories to make them more TS-friendly and avoid repeating too much code!

I have a few remarks on package.json mostly.

Also I think the things I fixed in https://github.com/opendatasoft/ods-dataviz-sdk/pull/110 apply to `visualizations-react` as well (output dir for UMD, paths to update in `package.json`)

packages/visualizations-react/package.json
etienneburdet: No it's a merge problem, they've been there back and forth with eslint 👌

packages/visualizations-react/package.json
etienneburdet: Actually I'm of the opinion of starting with an up-to-date rollup since it doesn't seem to break anything to upgrade 👌

packages/visualizations-react/package.json
etienneburdet: It doesn't seem we're using it anywhere though… just removing it doesn't break anything apparently 🤷

richterb: Let's remove it then, one less thing to worry about 😄

packages/visualizations-react/rollup.config.js
etienneburdet: For something like observable maybe? I'm not sure either… I don't see a reason to bundle to ES since we can import already. Maybe we don't bundle at all after all 😅 Same question could be asked for visualization…

richterb: `visualization` can be used in contexts such as browsers by just using a script tag and running `new Chart()`, so UMD makes sense here IMO (also, Observable indeed). But `visualization-react` requires `React` and I'm not sure you can do a `React` app using script tags, or in Observable. But let's keep it for now I guess...

packages/visualizations-react/stories/Chart/Formatting/LongDataLabels.stories.tsx
```diff
@@ -92,14 +92,16 @@ const pieDataFrame = [
...
+export const PieLongDataLabels = ChartTemplate.bind({});
+const PieLongDataLabelrsArgs: Props<DataFrame, ChartOptions> = {
     data: {
         loading: false,
         value: pieDataFrame,
     },
     options: {
         labelColumn: 'x',
         source: defaultSource,
+        // source: defaultSource,
```
wmai, line 104: A small oversight to remove.

packages/visualizations-react/.eslintrc.js
```diff
@@ -1,8 +1,107 @@
...
+        'react/react-in-jsx-scope': 'off',
+        'import/no-named-as-default': 'off',
+        // FIXME: With the introduction of TS, these rules cause linter errors (many of them quite justified). For the
+        // ... moment, they have been disabled but in the future we could re-enable them and fix the different errors.
+        '@typescript-eslint/naming-convention': 'off',
+        '@typescript-eslint/default-param-last': 'off',
+        '@typescript-eslint/no-var-requires': 'off',
+        // avoid the usage of test.only
+        'no-only-tests/no-only-tests': 'error',
+        "react/function-component-definition": [0, { namedComponents: ["function-expression", "arrow-function"], unnamedComponents: "arrow-function" }],
```
wmai, line 105: Why not leave the default value for this rule? Personal preference? (If yes, it's fine with me, I'm asking mostly out of curiosity ^^). The definition of this rule in the platform is:
```
"react/function-component-definition": [0, { namedComponents: 'function-expression', unnamedComponents: 'function-expression'}],
```

packages/visualizations-react/.eslintrc.js
```diff
@@ -1,8 +1,107 @@
...
+        'no-underscore-dangle': 0,
+        'react/jsx-one-expression-per-line': 0,
+        'arrow-parens': [2, 'as-needed', { requireForBlockBody: true }],
+        'react/jsx-props-no-spreading': 0,
+        'jsx-a11y/label-has-for': 0,
+        'jsx-a11y/label-has-associated-control': ['error', {
+            labelComponents: ['label'],
+            controlComponents: ['input', 'CheckboxInput', 'RadioInput', 'SelectInput', 'TextInput', 'SwitchInput'],
+            depth: 25,
+        }],
```
wmai, line 88: This rule seems to be very related to the platform code, I wonder if it is relevant to define it here. It also names components that are only present in the platform. Such as `CheckboxInput`, `RadioInput`, `SelectInput`, `TextInput` and `SwitchInput`.

wmai (commented):
Great job and thank you so much for taking on this chore!

I only have a few questions about the ESLint configuration but nothing blocking to merge IMO.

packages/visualizations-react/.eslintrc.js
etienneburdet: I think there was a slight looping problem with Prettier here, but yes personal preference also, I really don't like function declaration for components (it made stories really strange) and I really prefer arrow for anything unamed/inlined. But that's pretty open to discussion…

packages/visualizations-react/.eslintrc.js
etienneburdet: Yup, pure copy-paste, I'll have to rework that indeed. I'm not sure we need it at all, a11y should be done in `visualization` mainly I guess.

packages/visualizations-react/stories/Chart/Formatting/LongDataLabels.stories.tsx
richterb: Fixed here 8f65f9865ef34a950a8985836ed2c5300c5c32fd

packages/visualizations-react/.eslintrc.js
richterb: Agreed, removed here 766042c1e7a4e7d0d0096a7f52bc737611226797
Note that if we release filters somewhere, we may want a similar rule again to make sure we don't have a text input without a label for example

packages/visualizations-react/.eslintrc.js
richterb: As I don't have any opinion on the subject, and in the spirit of running ahead, I suggest we don't touch anything for now and go with this 😅

PR #114, opened 2022-09-26
chore: make everything node16

etienneburdet:
## Summary

The goal for this PR is to make everything run on node 16. 

In local dev should work with `nvm use, nvm use 16, nvm lts/gallium`. 

CI should run on 16 too. It should also prevent crashes due to using node 18, which
### Changes
I added a root .nvmrc, changed version in all CI scrpits and specified <18 for engine in packages.json.

### Changelog
"Added utilities to enforce the use of node 16 everywhere".

## Open discussion
We could specify 16.7.1 for CI, I'm not sure what it would change.

## To be tested
CI, please be green 🤌 

- [x] Description is complete


richterb:
I think at least the tests should run on 14 as well (test coverage and linting have no reason to work on 16 and not on 14, but regular tests may detect real issues). After all, the main user of the SDK (us) is still on Node 14 😅

.github/workflows/api-client-coverage.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        node: ['14.x']
+        node: ['16.x']
```
wmai, line 13: About @richterb remark, I wonder if we could use the following config in order to make the tests run on multiple NodeJS versions:

```
node-version: ['14.x', '16.x']
```

richterb: As far as I can tell from the doc, yes this should work, everything has already been written in this action to allow for it, by using `matrix` (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) and `matrix.node` as vars in the steps.

richterb:
The tests are now running on both Node 14 and 16 🎉

PR #121, opened 2022-10-14
Use TurboSnap feature from Chromatic to reduce quota usage

richterb:
## Summary

The goal for this PR is to reduce our Chromatic quota usage by leveraging TurboSnap, which should only compare snapshots for code that has changed.

### Changes

See https://www.chromatic.com/docs/turbosnap#enable

The PR just adds the relevant option to Chromatic.

## Review checklist

- [ ] Description is complete

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -33,3 +33,4 @@ jobs:
           projectToken: ${{ secrets.CHROMATIC_TOKEN }}
           workingDir: ./packages/visualizations-react
           storybookBuildDir: storybook-static
+          onlyChanged: true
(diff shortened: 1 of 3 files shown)


richterb:
Well, unfortunately, TurboSnap won't work for us :( https://www.chromatic.com/docs/turbosnap#specify-external-files-to-trigger-a-full-re-test-when-they-change

Basically, it can only detect direct dependencies to stories. In our case, most changed we want to detect happen in `visualization`, and these changes are imported through node_modules and Webpack can't trace beyond it. This means that changes are not detected. And if we add these files with `--externals`, this mean any change in there will trigger a full re-test, which means TurboSnap is useless here.

PR #122, opened 2022-10-20
Split choropleth components

RafaelSzmarowski:
## Summary

The goal for this PR is to split GeoJson and Vector Tiles Choropleths into two distinct components. 

### Changes

- Choropleth is split into ChoroplethGeoJson and ChoroplethVectorTiles
- MapRender remains unique for both components
- ColorsScale is renamed ColorScale everywhere 
- Most of the Choropleth computing functions have been moved to utils

## Open discussion

- Remove MaxZoom and pass it as an option

## To be tested

No regression ?

## Review checklist

- [ ] Description is complete


packages/visualizations/src/components/Map/types.ts
```diff
@@ -37,11 +37,11 @@ export interface MapFilter {
     value: string | string[];
 }
 
-export interface GeoJsonChoroplethOptions extends ChoroplethOptions {
+export interface ChoroplethGeoJsonOptions extends ChoroplethOptions {
```
richterb, line 40: One of the "easy" benefit of splitting IMO would be to not require this `shapes` object from the outside.

We could have directly `shapes` expecting a `GeoJSON` here, and for `ChoroplethVectorTilesOptions` directly have a `shapesTiles` property (a different name for clarity) expecting `ChoroplethShapeVectorTilesValue` but without the `type` anymore.

Then `ChoroplethGeoJson.svelte` and its VTiles counterpart have already the code that handles it, so it shouldn't make things too different. From the outside's point of view, using each of these components will be much easier to understand.

packages/visualizations/src/components/Map/utils.ts
```diff
@@ -227,8 +229,78 @@ export const getFixedTooltips = (
...
+    /** Transform a filter object from the options into a Maplibre filter expression */
+    const { key, value } = filterConfig;
+    const filterMatchExpression: string[] = ['in', key];
+    (Array.isArray(value) ? value : [value]).forEach((filterValue) => {
+        filterMatchExpression.push(filterValue.toString());
+    });
+    return filterMatchExpression;
+};
+
+export const defaultFormat: ChoroplethTooltipFormatter = ({ value, label }) =>
```
richterb, line 242: Maybe `defaultTooltipFormat` to be more explicit?

packages/visualizations/src/components/Map/utils.ts
```diff
@@ -227,8 +229,78 @@ export const getFixedTooltips = (
...
+    type: 'fill',
+    layout: {},
+    paint: {
+        'fill-color': fillColor,
+        'fill-opacity': 0.8,
+        'fill-outline-color': DefaultColor,
+    },
+});
+
+export const computeMatchExpression = (
```
richterb, line 297: I wonder if we could enforce the `FilterSpecification` from `maplibre` for the return value here. At first glance it seems to be intended for it. If it works, it would make further changes in this function more "secure".

packages/visualizations/src/components/Map/utils.ts
RafaelSzmarowski: 👍

packages/visualizations/src/components/Map/utils.ts
RafaelSzmarowski: Yes Indeed, I've added it and it seems working

packages/visualizations/src/components/Map/WebGl/ChoroplethGeoJson.svelte
```diff
@@ -0,0 +1,129 @@
...
+    let colorScale: ColorScale;
+
+    let aspectRatio: number | undefined;
+    let renderTooltip: MapRenderTooltipFunction;
+    let bbox: BBox | undefined;
+    let activeShapes: string[] | undefined;
+    let interactive: boolean;
+    let legend: MapLegend | undefined;
+    let filter: MapFilter | undefined;
+    let filterExpression: (string | string[])[] | undefined;
```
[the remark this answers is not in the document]
RafaelSzmarowski: Ok ! I just removed the filter related code lines from the geojson choropleth and made it by default to null in `MapRender`. Do you think it's enough or we should try to be more explicit that it's not supported in GeoJson choropleths ?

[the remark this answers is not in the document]
richterb: I think we just need to make sure `filter` does not appear in the type used to describe the options. I guess we just need to remove it from `ChoroplethOptions` and only include it in the VectorTiles specific variant of the type.

packages/visualizations-react/src/components/ChoroplethGeoJson.tsx
```diff
@@ -1,12 +1,12 @@
-import { FC } from 'react';
 import {
-    DataFrame,
-    SvgChoropleth as _SvgChoropleth,
+    ChoroplethGeoJson as _ChoroplethGeoJson,
     ChoroplethOptions,
```
richterb, line 3: Each React wrapper can now return exactly the specific type (`ChoroplethGeoJsonOptions`, `ChoroplethVectorTilesOptions`)

etienneburdet (commented):
I did quick look around of the code only, everything seems in its place and the new utils functions are well split 👍 

I'll do some real-life testing on the studio side, but it can be merged meanwhile if need be 🤷

packages/visualizations/src/components/Map/types.ts
```diff
@@ -128,6 +111,13 @@ export interface ChoroplethFixedTooltipDescription {
 
 export type MapRenderTooltipFunction = DebouncedFunc<(f: Feature) => string>;
 
+export type ComputeTooltipFunction = (
+    f: Feature,
```
wmai, line 115: I'm nitpicking, sorry, but could we name the parameter `feature` to be more explicit?

packages/visualizations/src/components/Map/types.ts
```diff
@@ -95,31 +95,14 @@ export interface ChoroplethDataValue {
     label?: string;
 }
 
-export enum ChoroplethShapeTypes {
-    GeoJson = 'geojson',
-    VectorTiles = 'vtiles',
-}
-
-/** `ChoroplethShapeValue` implementation based on a GeoJSON FeatureCollection.  */
-export interface ChoroplethShapeGeoJsonValue {
```
wmai, line 104: In order to use it oustide, I feel like we still need to export a type representing this object. Something like:

``` TypeScript
export type ChoroplethShapeGeoJsonValue = FeatureCollection;
```

packages/visualizations/src/components/Map/types.ts
```diff
@@ -95,31 +95,14 @@ export interface ChoroplethDataValue {
     label?: string;
 }
 
-export enum ChoroplethShapeTypes {
-    GeoJson = 'geojson',
-    VectorTiles = 'vtiles',
-}
-
-/** `ChoroplethShapeValue` implementation based on a GeoJSON FeatureCollection.  */
```
wmai, line 103: The type `ChoroplethShapeValue` doesn't exist anymore. This comment is obsolete.

packages/visualizations/src/components/Map/types.ts
richterb: `ChoroplethGeoJsonOptions` already defines that it expects a `FeatureCollection` for `shapes`, which seems fine to me: I think it's clearer to talk about a well-known thing like GeoJSON feature collection, than something named by us which is equivalent.

richterb (approved):
Seems good to me!

I didn't test installing and using it from the platform, so we may miss something, but I definitely prefer this approach

packages/visualizations/src/components/Map/types.ts
wmai: Indeed, but `FeatureCollection ` comes from `@types/geojson` which may not be installed in projets that may use our SDK. Plus, I feel like we also need to export it with our naming convention. But I don't have a strong opinion on this subject.

richterb: But if we create a new type that is based on `@types/geojson`'s `FeatureCollection`, the projects will still need to have the package to resolve the type anyway. But indeed that's something I never thought about, how does it work in that case... Maybe it just does complain about the missing package anyway. And maybe there's a way to directly inline and package all the types in the definitions that are built (the `*.d.ts` files). It's worth having a look at the best practices when we have the time.

Regardng our naming conventions, I feel like it's you're installing the SDK, and having a look at the expected parameters, having "`shapes: FeatureCollection`" tells you immediately what you need and makes it simple. I don't think there's an added value to add our own named version that requires the developer to click one step further and understand that we're expecting GeoJSON, since there's really no nuance to it, it's really just a FeatureCollection without any other requirement. But it's worth a deeper philosophical discussion on a wednesday morning :)

wmai:
> I didn't test installing and using it from the platform, so we may miss something, but I definitely prefer this approach

@richterb I've tested it and since some interfaces names have changed, a PR renaming all of those will be needed on the platform.

packages/visualizations/src/components/Map/types.ts
etienneburdet: I really prefer using `FeatureCollection` as it is. It's more generic, more explicit and it doesn't induce the idea that there might be a difference.
The package isn't a problem IMO, it's in the dev dependencies anyways and only will only be useful for devs. So it will be installed by `npm i`.

packages/visualizations/src/components/Map/types.ts
RafaelSzmarowski: Ok 👍 I've just changed it !

packages/visualizations/src/components/Map/types.ts
RafaelSzmarowski: Thanks, I totally forgot about that one !

PR #128, opened 2022-11-16
Add Navigable map component

etienneburdet:
## Summary

The goal for this PR is to add a component with minimaps, where clicking on the minimap will center the main map on its shape. The manimap should color with choropleth data.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-36556](https://app.shortcut.com/opendatasoft/story/36556/sdk-as-an-ods-dev-i-have-a-sdk-component-that-can-render-french-drom-com-next-to-a-standard-map-of-france-in-order-to-show).

### Changes
Added a new component that combine SVG choropleth and vector choropleth, with button callbacks.

I also had to introduce the a new `viewBox` prop, to differenciate from the `boundingBox`: the former is set the view, but you can zoom/pan out of it, the later constrains the view. I think we want to be able to manually zoom out when clicking on a minimap.

Adde turf only for creating the story (to find the bbox of the geojson). Likewise, it's how some people would use it.

Added empty css selectors on map wrapper, buttons etc. to be able to style from the outside.

As of this PR, only SVG maps are available as buttons, the "back" button is only stylable with css and clicking the button will only center the map on a given bbox.

### Changelog
 "A new component is available, that allows to navigate a choropleth map with minimaps (e.g. for oversea territories)"

## Open discussion
So, this one is very "Studio" and rigid, but I feel that's enough code for a first pass, but leaves a few question open.

The first point to be discussed is styling: I decided to add very simple classes, instead of custom properties like in the KPI. It seemed to be much easier to handle and reduce API surface. Should we do the same in the KPIs?

As for the callback, I went the way of adding a prop to watch and we could add many more if needed. This would limit the user to what we want. Another way of doing that would be to bind a public API (or even the `map` instance itself) to the `MapRender` component like this. This would allow a dev to use all (public) methods. Another option would be to take callback of the form: `mapCallback = map =>  args => { map.doStuff(); … }` and then do `callback = mapCallback(map)` inside the mapRender.

And the come the buttons, both the back button and the minimap. I tried to accept children or even HTML as props, but without much success. Ideally, I would like something like: 
```
<NavigableMap>
  <div slot="main">
    <MyMap />
  </div>
  <div slot="minimaps">{ loop mini maps}</div>
  <div slot="back-button">…</div>
</NavigableMap>
```
Which could be done with components as props too, I think. But it's far fetched in our current implementation.

## To be tested
Test with and without `bbox`/`viewbox`, maybe with `null` and `0` in some places.

## Review checklist

- [x] Description is complete


packages/visualizations/src/components/Map/WebGl/ChoroplethVectorTiles.svelte
```diff
@@ -1,5 +1,3 @@
-<svelte:options immutable={true} />
```
richterb, line 1: Why this change here and in MapRender? We had a look at this one year ago I think, and we opted for this because of our use of a global `options` parameter containing an object. Without `immutable`, Svelte will consider that the object will have changed at every render. Overall this only has drawbacks, and usually if removing it seems to fix something, it's probably a symptom of something broken coming from our own React code (typically mutating the `options` object instead of creating a new one).

For reference:
- [the equality check difference between the two modes is defined here](https://github.com/sveltejs/svelte/blob/master/src/runtime/internal/utils.ts)
- [the two functions](https://github.com/sveltejs/svelte/blob/master/src/runtime/internal/utils.ts) (the `a!= a` is there to detect `NaN`, the rest is rather clear)

richterb:
The NavigableMap story in Storybook seems broken:
```
Error: sources.shape-source-815.minzoom: number expected, undefined found
    wn maplibre-gl.js:31
    Jt maplibre-gl.js:35
    _validate maplibre-gl.js:35
    addSource maplibre-gl.js:35
    addSource maplibre-gl.js:35
    updateSourceAndLayer MapRender.svelte:205
    update MapRender.svelte:228
    update index.mjs:1089
    flush index.mjs:1060
```

packages/visualizations/src/components/Map/Combinations/MiniMap.svelte
```diff
@@ -0,0 +1,38 @@
...
+    <button type="button" class:active on:click>
+        <SvgChoropleth {data} {options} />
+    </button>
+{/if}
+
+<style>
+    button {
+        cursor: pointer; /* Override user agent stylesheet */
+        height: 100%;
+        flex: 1;
```
richterb, line 33: I suppose it's a rest of a previous implementation, but right now this prop is obsolete because the container is grid not flex

packages/visualizations/src/components/Map/Combinations/MiniMap.svelte
```diff
@@ -0,0 +1,38 @@
...
+    export let data: { value: ChoroplethDataValue[] };
+    export let map: NavigationMap;
+    export let colorScale: ColorScale | undefined;
+
+    $: ({ label } = map);
+    $: options = { shapes: map.shapes, colorScale };
+</script>
+
+{#if showTooltip}
+    <Tooltip {label}>
```
richterb, line 18: In React we could have mutualized the button code in a separate component declared inline the same file, so that it would look like
```svelte
{#if showTooltip}
    <Tooltip {label}>
        {minimapButton}
    </Tooltip>
{:else}
    {minimapButton}
{/if}
```

Is there a clean way to do something similar in Svelte (without creating an entire separate component file?)

It would be a bit cleaner and prevent future mistakes where we edit one but not the other, but maybe not at the price of the overhead of an entire separate component...

packages/visualizations/src/components/Map/Combinations/NavigableMap.svelte
```diff
@@ -0,0 +1,85 @@
...
+    const defaultInteractive = true;
+    $: ({ interactive = defaultInteractive } = options);
+
+    let viewBox: BBox | undefined;
+    $: viewBox = bbox;
+
+    let active: number | undefined;
+    let height: number;
+
+    const setViewBox = (map: NavigationMap, i: number) => () => {
```
richterb, line 23: The `interactive` prop should also block clicks on the button. Ideally it should also remove the hover effect as well... Worst case, a good old `pointer-events: none;` would block all that, although a bit roughly (and imperfectly in terms of accessibility)

packages/visualizations/src/components/Map/Combinations/Tooltip.svelte
```diff
@@ -0,0 +1,49 @@
...
+
+    function hideTooltip() {
+        hovered = false;
+    }
+</script>
+
+<div
+    class="tooltip-container"
+    bind:this={containerElement}
+    on:mouseover={showTooltip}
```
richterb, line 22: I don't know if we ever discussed mobile here with the Product Design team. The tooltip isn't displayed (although clicking repeatedly in Firefox's mobile testing seems to sometimes trigger it), and I'm not sure what we want to do here. Otherwise, mobile users will never have a label to describe the buttons...

packages/visualizations/src/components/Map/Combinations/Tooltip.svelte
```diff
@@ -0,0 +1,49 @@
+<script lang="ts">
+    export let label = '';
+    let containerElement: HTMLDivElement;
+    let x: number;
+    let y: number;
+    let hovered = false;
+
+    function showTooltip() {
```
richterb, line 8: Why not tippy? I supposed the tooltip we have on KPIs looked similar to what we expect here, and we'd have visual consistency and maybe some goodies we don't think of (accessibility like showing on focus or stuff like that)

packages/visualizations/src/components/Map/Combinations/NavigableMap.svelte
```diff
@@ -0,0 +1,85 @@
...
+        active = undefined;
+        viewBox = bbox;
+    };
+
+    $: vectorOptions = { ...mainOptions, colorScale, viewBox };
+    /* Number of column x largest colmuns.
+     * The maps will then auto-adjust between 52 and 72 to fit the best
+     * It's stable and doesn't involves too much JS computaton either.
+     */
+    $: navMapsWidth = height ? Math.ceil((navigationMaps.length * 72) / height) * 72 : 72;
```
richterb, line 38: In my tests (Firefox & Chrome), it seems the buttons are always at 72x72. Even when playing with the browser size etc. I'm not sure if something's missing.

As a result, since the buttons never take all their space, their block doesn't appear centered so it looks a bit weird.

packages/visualizations/src/components/Map/types.ts
```diff
@@ -24,8 +24,15 @@ export interface ChoroplethOptions {
          * By default, the label will be taken from a `label` property in the shapes if it exists, or fallback to the key used to map the data and shapes. */
         labelMatcher?: ChoroplethTooltipMatchers;
     };
-    /** Initial position of the map when rendering. If undefined, with GeoJSON shapes the map will automatically zoom to fit all content, and on VTiles map, it will display the world. */
+    /** Maximum boundaries of the map, outside of which the user cannot zoom/move */
     bbox?: BBox | undefined;
+    /** Position of the map when rendering.
+     * If undefined, will default, in order to:
+     * - bbox if specified
+     * - Fit of teh GeoJson with a geojson soruce
```
richterb, line 32:
```suggestion
     * - Fit the content if the source is GeoJSON
```

packages/visualizations/src/components/Map/types.ts
```diff
@@ -47,6 +54,18 @@ export interface ChoroplethVectorTilesOptions extends ChoroplethOptions {
...
+export type NavigationMap = {
+    shapes: FeatureCollection;
+    bbox: BBox;
+    label: string;
+};
+
+export type NavigableChoroplethOptions = ChoroplethVectorTilesOptions & {
+    navigationMaps: NavigationMap[];
+    /* Miniman container size below which the buttons will be on the right */
+    horizontal?: number;
```
richterb, line 66: I think this is no longer useful now (it's not used by the current code)

richterb (commented):
I didn't spend too much time on the responsive behavior since it's not entirely over, but I reviewed the rest of the code. This is a huge lot of work 👏

packages/visualizations/src/components/Map/Combinations/MiniMap.svelte
etienneburdet: As far as I know, it has to be a separate component. But we shouldn't be afraid of doing a lot of small components IMO. I can do that!

packages/visualizations/src/components/Map/Combinations/Tooltip.svelte
etienneburdet: I think there is a better event to be used `on:touch` or something like that, that cover both hover and touch. I'll dig.

PR #130, opened 2022-12-01
chore: added a command to rebuild lock files

etienneburdet:
## Summary
This PR adds an npm script to rebuild package lock if they happen to not be in sync

With the new Lerna version, they are not rebuilt—which is great because they used to f* up the dff—and so can be out of sync with the package.json. Running the command will solve that.


richterb (approved):
Thanks 🙏

PR #131, opened 2022-12-02
Feat: Rename `tooltip.labelFormatter` callback into `tooltip.formatter` in Choropleths options

wmai:
## Summary
The goal for this PR is to rename the ambiguous `tooltip.labelFormatter` callback into `tooltip.formatter` which in fact can format the label AND the value for Choropleths tooltip.

_(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-37317](https://app.shortcut.com/opendatasoft/story/37317)._

## Review checklist
- [x] Description is complete

### DIFF
--- a/packages/visualizations/src/components/Map/utils.ts
+++ b/packages/visualizations/src/components/Map/utils.ts
@@ -269,7 +269,7 @@ export const computeTooltip: ComputeTooltipFunction = (
     }
 
     const tooltipRawValues: TooltipParams = {
-        value: matchedFeature?.y.toString(),
+        value: matchedFeature?.y,
         label: tooltipLabel,
         key: hoveredFeature.properties?.[matchKey], // === matchedFeature.x
     };
--- a/packages/visualizations/src/components/Map/utils.ts
+++ b/packages/visualizations/src/components/Map/utils.ts
@@ -17,7 +17,7 @@ import type {
     ComputeTooltipFunction,
     ChoroplethLayer,
 } from './types';
-import { ChoroplethTooltipMatcherTypes } from './types';
+import { ChoroplethTooltipMatcherTypes, TooltipParams } from './types';
 
 export const LIGHT_GREY: Color = '#CBD2DB';
 export const DARK_GREY: Color = '#515457';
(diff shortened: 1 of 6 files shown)


etienneburdet (approved):
LGTM! 🏂

etienneburdet (approved):
LGTM! 🏂 Sorry for taking that long after the SDK part… just forgot about it. The new enum is very clean and readable. 

Also, just maybe add to make an SDK release, next to the migration script, so we don't forget.

packages/visualizations/src/components/Map/utils.ts
```diff
@@ -268,16 +268,12 @@ export const computeTooltip: ComputeTooltipFunction = (
...
     }
 
-    const tooltipRawValues: {
-        value?: number;
-        label: string;
-        key: string;
-    } = {
-        value: matchedFeature?.y,
+    const tooltipRawValues: TooltipParams = {
+        value: matchedFeature?.y.toString(),
```
richterb, line 272: It's tied to changes on the Studio side, but I'm not sure why here, where we know we start from a number, we have to transform it into a string, then pass that string to `Intl.NumberFormat` which expects a number. Doing number -> string -> number in the middle is a risk of losing information (for example truncating values in very long floats), and I'm not sure what the gain is. Also it seems logical to expect a value that is a number here, in terms of function signature / API, because in the context of a choropleth it will always be a number underneath.

richterb (commented):
A fundamental question on typings 😅

packages/visualizations/src/components/Map/utils.ts
wmai: Indeed it is a very good remark, thank you. I must admit I didn't think much about it when I decided to change the type of the `TooltipParams` interface. Fixed [in this branch](https://github.com/opendatasoft/ods-dataviz-sdk/pull/131/commits/6d6f829ca3a06d566ba75303d090996e7d94f941) and in the [Studio as well](https://github.com/opendatasoft/platform/pull/9903/commits/f78847a8d16ca5604a116f7d17982eed7a1ae190).

richterb (approved):

PR #134, opened 2022-12-08
Poc-filtre

etienneburdet:
So… here is my take on filters. It's very minimal and add hoc for now, but hopefully the idea is there.

## Basic Idea
The basic idea is to add a "primitive" to the SDK, which represents filterable data, that is reactive across all of our blocks and can be mutated by others, like filter here. (Other primitives could be colorscale, condition for exemple). It's a svelte store, so that it's reactive inside all of our components. 

As you can see, you can totally use a svelte store in react (because it's pure JS) as long as it's passed to our components. It allows us to pass "filtered data" instead of data.

Right now it only does the "loading true/false" part as a bonus, but we could add: 
* Storing initial value and reseting to default.
* Same data could be modified by multiple components, we could track them if need be.
* Read the value outside of the component. One can subscribe to a Svelte store anywhere, see [this hook](https://stackoverflow.com/a/64320752/7508148) for exemple.

Everything would always be reactive in all the components.

## Not implemented
It's not typed yet, but the idea would it's easy to ensure a correct data type.

Components accepting both store and plain value. As you can see, I hard coded the store in the KPI, but in the end we should accept both format I think. There many ways to do that, either by accepting two dinstinct props or converting depending on type (with assertion).

## Questions
Right now, this new store/observable etc. is not specific to the api-client, but there is an argument for to be. If so my guess is we could move quite a bit of the `useDataFrame` into it. Or prodive to versions or provide an ods speicifc fetcher… whatever, I think there is room reduce boilerplate code for someone who would like to add

### DIFF
--- a/packages/visualizations-react/package-lock.json
+++ b/packages/visualizations-react/package-lock.json
@@ -17,6 +17,7 @@
                 "@babel/preset-env": "^7.19.0",
                 "@babel/preset-react": "^7.18.6",
                 "@babel/preset-typescript": "^7.18.6",
+                "@opendatasoft/api-client": "^0.5.0",
                 "@rollup/plugin-babel": "^5.3.0",
                 "@rollup/plugin-commonjs": "^19.0.0",
                 "@rollup/plugin-json": "^4.1.0",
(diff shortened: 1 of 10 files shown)


etienneburdet (commented):
So I added a second filter (on another field) and clear function (it arguably be the filter function).

I also added a clear button that demonstrate what I wanted to: it's possible to clear one specific value from anywhere in the app, natively.

The little factory that return the filter and clear factory is specific to ods data, but could really be an helper function in the api-client.

But the whole thing can work with any function actually. For exemple the filter function could just use `.filter` on an in memory array, fetch from somewhere else etc.

PR #136, opened 2022-12-15
POC of a post-brainstorm filter implementation

richterb:
Implementation of what we discussed after our previous POCs. The exact packaging of this (separate packages etc) to be discussed.


etienneburdet (commented):
Just dropped my personal vision for later discussion (before I forget about what I wanted to do). Did it as a review, for practical reasons, but i's more for leaving a trace than anything else 😛 Very tied to #134 in React story as an exemple implem.

PR #139, opened 2023-01-04
Basic filter implementation

richterb:
WIP


etienneburdet (approved):
Just added two base stories, for good measure.

I also removed the `$: applyFilter(value)`, in favor of dispatching the event on the domEvent itself. It shouldn't change anything.

Otherwise, LGTM! We can certainly improves the Implementations/wrappers when we accepts any kind of props, but that's for later!

packages/visualizations/src/filters/components/Select/Select.svelte
```diff
@@ -13,25 +13,21 @@
...
-    $: applyFilter(selected);
 </script>
 
 <div>
-    <select bind:value={selected}>
-    {#each availableValues as availableValue}
-        <option value={availableValue}>{availableValue}</option>
-    {/each}
+    <select on:change={(e) => applyFilter(e.currentTarget.value)}>
+        <option value={null}>Select a value</option>
```
wmai: @etienneburdet Changing `value={null}>` to `value>` because the value will be stringified into `"null"` (a990143)

PR #140, opened 2023-01-05
Add navigation mini maps svg buttons 

RafaelSzmarowski:
## Summary

The goal for this PR is to add minimap buttons to the Choropleth both geojson and vector tiles, where clicking on the minimap will center the main map on its shape. The minimap should color with choropleth data.

⚠️ This branch is a fork and a rework of this previous opened PR: [opened PR](https://github.com/opendatasoft/ods-dataviz-sdk/pull/128)
The main cause for the rework is an attempt to make the whole structure lighter. In the previous PR, a NavigableMap component was combining the vector tiles choropleth and the MiniMaps based on our svg choropleth component. It appears that the MiniMaps button could be passed as a prop directly to both our choropleth vector tiles and geojson which seems to make code more factorized.

As for now the buttons will always be displayed below legend, further investigations will be made to decide if a horizontal or vertical layout of the buttons is needed.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-36556](https://app.shortcut.com/opendatasoft/story/36556/sdk-as-an-ods-dev-i-have-a-sdk-component-that-can-render-french-drom-com-next-to-a-standard-map-of-france-in-order-to-show).

### Changes

- New MiniMap component based on rendering svg choropleths
- Choropleth Geojson and Vector Tiles can now receive a new props navigationMaps that renders navigation mini map buttons.
- The rendering of the MiniMaps is common for both of the choropleth and located in MapRender
- New viewBox value along bbox as the displayed Map can now be naigable
- Max bounds are not set if the map is navigable
- new BackButton component to handle reset viewBox
- new storybook stories

## Review checklist

- [ ] Description is complete


packages/visualizations/src/components/Map/WebGl/MapRender.svelte
```diff
@@ -125,8 +146,10 @@
         resizer = new ResizeObserver(
             debounce(() => {
                 map.resize();
-                if (mapReady && bbox) {
-                    fitMapToBbox(bbox);
+                if (mapReady && viewBox && navigationMaps) {
```
richterb, line 149: It's a bit complicated between `bbox` and `viewBox` now I think. At some point I documented in the types so that the difference is clear, and I think we shouldn't do implicit logic here, and rather follow the documentation:
```
display the world. */
    /** Maximum boundaries of the map, outside of which the user cannot zoom/move */
    bbox?: BBox | undefined;
    /** Position of the map when rendering.
     * If undefined, will default, in order to:
     * - bbox if specified
     * - Fit the content if the source is GeoJSON
     * - The world
     */
    viewBox?: BBox | undefined;
```

So, it's up to the developer (or in our ase, the Studio) to pass the right stuff: for us, it's either:
- no viewBox, and a bbox (vtiles) or none (geojson)
- a viewBox, and no bbox (navigable maps)

You could imagine a situation where you'd pass the bbox of a country, and a smaller maps to zoom to specific territories inside it. Either way, it would be predictable. What do you think?

richterb: Also it would probably simplify a lot of code there, e.g. `setViewBox` and `setMapBounds` will both only rely to respectively `viewBox` and `bbox`, which is easier to follow

RafaelSzmarowski: Indeed, I went a bit fast and I’ve missed this logic and thought we would like to only rely on viewBox with navigable maps.

I tried to fix this, so if there is a bbox it will only set the max bounds of the map and as you well described it in the documentation, the viewBox will take in this order custom values or bbox or fit content in GeoJson or the world bbox. It is responsible of only the displayed view location on the map. Hope I haven’t missed something but it seemed that in storybook it works as expected.

packages/visualizations/src/components/utils/MiniMap.svelte
```diff
@@ -0,0 +1,42 @@
...
+
+    $: ({ label } = map);
+    $: options = { shapes: map.shapes, colorScale };
+</script>
+
+{#if showTooltip}
+    <button
+        type="button"
+        class:active
+        on:click
```
KevinFabre-ods, line 21: `on:click` is unused ? To delete ?

packages/visualizations/src/components/utils/MiniMap.svelte
```diff
@@ -0,0 +1,42 @@
...
+        use:tippy={{
+            content: label,
+        }}
+    >
+        <SvgChoropleth {data} {options} />
+    </button>
+{:else}
+    <button type="button" class:active on:click>
```
KevinFabre-ods, line 29: `on:click` is unused ? To delete ?

packages/visualizations/src/components/utils/BackButton.svelte
```diff
@@ -0,0 +1,21 @@
+<script>
+</script>
+
+<button class="back-button" type="button" on:click>
```
KevinFabre-ods, line 4: `on:click` is unused ? To delete ?

KevinFabre-ods (commented):
For me, it was more a discovery of the code than a functional review. 

I just have a remark about the use of `on:click` events

packages/visualizations/src/components/utils/MiniMap.svelte
RafaelSzmarowski: I guess it's useful as it allows to pass a callback to the component. Without it doesn't seem to work.

packages/visualizations/src/components/utils/MiniMap.svelte
RafaelSzmarowski: Same as above, as far as I've tested it without all this on:clik it doesn't forward the on:click passed to the component I guess

packages/visualizations/src/components/utils/BackButton.svelte
RafaelSzmarowski: Same as above, as far as I've tested it without all this on:clik it doesn't forward the on:click passed to the component I guess

packages/visualizations/src/components/utils/MiniMap.svelte
KevinFabre-ods: I was too quick. 😅 
This is what you can find in the Svelte doc: 

> https://svelte.dev/docs#template-syntax-element-directives-on-eventname 
>
> If the on: directive is used without a value, the component will forward the event, meaning that a consumer of the component can listen for it.

packages/visualizations/src/components/utils/BackButton.svelte
```diff
@@ -0,0 +1,21 @@
...
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24"
+        ><path fill="none" d="M0 0h24v24H0z" /><path
+            d="M7.828 11H20v2H7.828l5.364 5.364-1.414 1.414L4 12l7.778-7.778 1.414 1.414z"
+        /></svg
+    >
```
etienneburdet, line 9: It's a bit ugly to have that hardcoded… we'll have to clean it at some point. But not now 😛

etienneburdet (commented):
Ok, so, I think it's the simplest way… for now 😛 The main goal of having the buttons and legend at the same level is here, we don't have to look further immediatly, indeed. And most importantly: it works!

Big scheme of thing, I think we can still extract many components, just made [a ticket for that](https://app.shortcut.com/opendatasoft/story/37950/brake-maprenderer-into-smaller-components-sdk).

PR #147, opened 2023-01-13
Fix viewbox resizer not on currentViewBox and max bounds not refreshing

RafaelSzmarowski:
## Summary

The goal for this PR is to fix two recent regressions : 

1) The resizer fits the map to the viewBox instead of the currentViewBox what produces a weird behaviour. If you select a map through navigation buttons, on resize it would take you back to default viewBox. Just replacing viewBox by the currentViewBox fixes that.

2) We removed setting the map max bounds from the resizer and as it is based on `map.getBounds()` after doing a first `fitBox(bbox)`, it causes a bad behaviour on resize and aspect ratio changes and sometimes constraining the map too much making it impossible to see everything.
The fix is to always set max bounds depending on the bbox or the currentBbox on each change, that way the map will be always limited even if you navigate through navigation buttons. It removes the necessity to have co-existing bbox and viewBox which IMO makes all the code easier to understand.

Please share your opinion as I'm not sure of what is the best behaviour to have 😄 

## Review checklist

- [ ] Description is complete

### DIFF
--- a/packages/visualizations/src/components/Map/WebGl/MapRender.svelte
+++ b/packages/visualizations/src/components/Map/WebGl/MapRender.svelte
@@ -37,7 +37,6 @@
     export let layer: MapLayer;
     // bounding box to start from, and restrict to it
     export let bbox: BBox | undefined;
-    export let viewBox: BBox | undefined = bbox;
     // option to disable map interactions
     export let interactive: boolean;
     // options to display legend
@@ -67,10 +66,6 @@
     let legendVariant: LegendVariant;
     $: legendVariant = clientWidth <= 375 ? 'fluid' : 'fixed';
 
-    // Used to set max bounds on map to avoid unecessary scrolls
-    let isViewBoxEqualToMap: boolean;
-    $: isViewBoxEqualToMap = bbox === viewBox;
-
     // Used to store fixed tooltips displayed on render
     // FIXME: This may not be useful anymore, and is very tied to Choropleth right now
     let fixedPopupsList: ChoroplethFixedTooltipDescription[] = [];
@@ -83,7 +78,7 @@
     let nav: NavigationControl;
 
     let mapReady = false;
-    let currentViewBox = viewBox;
+    let currentBbox = bbox;
     // Used to add a listener to resize map on container changes, canceled on destroy
     let resizer: ResizeObserver;
 
@@ -98,30 +93,22 @@
     const sourceId = `shape-source-${mapId}`;
     const layerId = `shape-layer-${mapId}`;
 
-    const fitBox = (box: BBox | LngLatBoundsLike) => {
-        // Using padding, keep enough room for controls (zoom) to make sure they don't hide anything
-        map.fitBounds(box as LngLatBoundsLike, {
-            animate: false,
-            padding: 40,
-        });
-    };
-
-    const setViewBox = (box: BBox | undefined) => {
+    const setBbox = (box: BBox | undefined) => {
         if (!box) {
             // zoom-out to bounds defined in the initialization
             map.setZoom(map.getMinZoom());
             return;
         }
-        fitBox(box);
-    };
-
-    function setMaxBounds(bounds: BBox) {
         // Cancel any saved max bounds to properly fitBounds
         map.setMaxBounds(null);
-        fitBox(bounds);
-        // Reset min zoom and movement
+        // Using padding, keep enough room for controls (zoom) to make sure they don't hide anything
+        map.fitBounds(box as LngLatBoundsLike, {
+            animate: false,
+            padding: 40,
+        });
+        // Set new map max bounds after bbox changes
         map.setMaxBounds(map.getBounds());
-    }
+    };
 
     function initializeMap() {
         const defaultCenter: LngLatLike = [3.5, 46];
@@ -149,18 +136,11 @@
 
     function initializeResizer() {
         // Set a resizeObserver to resize map on container size changes
-        // TODO: Do we really want to reset to the initial bbox each time?
         resizer = new ResizeObserver(
             debounce(() => {
                 map.resize();
-                if (isViewBoxEqualToMap) {
-                    map.setMaxBounds(null);
-                }
-                if (mapReady && currentViewBox) {
-                    setViewBox(currentViewBox);
-                }
-                if (isViewBoxEqualToMap) {
-                    map.setMaxBounds(map.getBounds());
+                if (mapReady && currentBbox) {
+                    setBbox(currentBbox);
                 }
             }, 100)
         );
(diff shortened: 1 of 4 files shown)


packages/visualizations/src/components/Map/WebGl/MapRender.svelte
```diff
@@ -149,12 +153,17 @@
         resizer = new ResizeObserver(
             debounce(() => {
                 map.resize();
-                if (mapReady && viewBox) {
-                    setViewBox(viewBox);
+                if (isViewBoxEqualToMap) {
+                    map.setMaxBounds(null);
```
richterb, line 157: I have a hard time reproducing a broken case with my data, but the `isViewBoxEqualToMap` means situations where we don't pass any viewBox, and it defaults to bbox right? Why is this situation special compared to specific viewbox and bbox? Maybe is it the other way around, and we only want to do this maxbounds gymnastic if there is already one (`map.getMaxBounds()`) at the moment of the resize?

richterb: Although that would be dangerous as well because you could have a bbox and then maxBounds that is very wide, and it would be reduced to the current viewBox suddenly, so no my logic is wrong 🙈

richterb: Any specific reason why we couldn't just setMaxBounds to the `bbox` option directly? There has to be a reason we didn't do this at the start, but I'm not sure why anymore. If so this would probably remove the need for all that.

RafaelSzmarowski: I wasn't precise enough on my PR description, it's because setting maxBounds is not only related to the bbox but more about the map canvas size... For example same bbox but different map aspect ratio results in different maxBounds

etienneburdet: I find the double conditions on `isViewBoxEqualToMap` a bit strange too.

IMO, we should not have to deal with maxBounds resizing, maxbound are the same whatever the zoom is. In my testing just:
```suggestion
                if (mapReady && currentViewBox) {
                    setViewBox(currentViewBox);
                }
```
seems to work unless I'm missing a case. We just fit the view on resize.

This was more something I forgot to remove in the first place with the introduction of the viewbox I think.

etienneburdet (changes requested):
I think there is an even simpler solution: set the maxBounds once and for all based on data and then only operate on viewBox.

This needs to be tested though, I might have missed some edge cases.

RafaelSzmarowski:
> I think there is an even simpler solution: set the maxBounds once and for all based on data and then only operate on viewBox.
> 
> This needs to be tested though, I might have missed some edge cases.

I think you miss the case where you change the aspect ratio of the map as the map bounds are defined based on the map canvas aspect ratio for a precise bbox and not only the bbox coordinates... It's quite hard to explain without a demo 😄 and in the France use case the bug was not appearing, it's easier to see that on the USA and Switzerland maps. 

The solution that appears to be the best is to refresh the maps bounds each time when you change the viewBox. It's a small change that will make the existence of two props bbox and viewBox obsolete but I think it makes things easier. I'll push it and let you have a look.

RafaelSzmarowski:
It should be easier to follow now, I have removed the double bbox and viewBox mechanism. The new mechanism is to always set a max bounds on map after the bbox is changing. Now in MapRender there is an initial bbox and the currentBbox which refers to the bbox that is currently displayed (for example when you navigate through the maps). I will change my PR description as well.

richterb (approved):

etienneburdet (commented):
I am of the opinion that we still need to distinguish between the "maximum boundaries" and where we want to place the camera, having only one param will limit us at some point. This is also the cause of the jitter when zooming out sometimes: if you zoom out of the boundaries (e.g. zooming out from a corner), the camera will shake before finding its place back.

One solution could be to set the min zoom when the data changes and/or on navigation with buttons, and then just fitBounds (not maxBounds). This would prevent from zooming out into the white, remove the maxBounds changing on resize and still allow to pan the map…

This can wait a further PR though, if we need the new countries out fast.

richterb:
> I am of the opinion that we still need to distinguish between the "maximum boundaries" and where we want to place the camera, having only one param will limit us at some point. This is also the cause of the jitter when zooming out sometimes: if you zoom out of the boundaries (e.g. zooming out from a corner), the camera will shake before finding its place back.
> 
> One solution could be to set the min zoom when the data changes and/or on navigation with buttons, and then just fitBounds (not maxBounds). This would prevent from zooming out into the white, remove the maxBounds changing on resize and still allow to pan the map…
> 
> This can wait a further PR though, if we need the new countries out fast.

Restricting the movement is especially useful/lifesaver (IMO) for one thing at the moment: when you're scrolling through the page, and your movement catches the map instead of the page scroll. If we block the max zoom, we avoid the issue when people are scrolling from a desktop using a mousewheel or a touchpad, however when on mobile, you're scrolling with touch, and here if you do the same gesture on the map, you'll pan the camera and not zoom out. Design is working on a way to prevent this stuff so hopefully it won't be an issue anymore in some time 🤞 

I agree we will probably need to independently drive the camera at some point in the future, however I'm not sure when or why (could be for marker maps but maybe we won't use the same MapRender; could be for the zoom-on-selected-shape but maybe we'll do it directly using events and not via a property; could be to pre-set a default view, but it's not planned for now and nobody ask for it -yet-, although I suspect it will come). I think this PR, as it is, solves our current issues with our current situation, and since it just has a single `bbox` property that won't change semantics later, when we re-introduce `viewBox` later, we'll have hands free to do exactly what we need.

packages/visualizations/src/components/Map/WebGl/MapRender.svelte
```diff
@@ -94,30 +93,22 @@
...
         map.setMaxBounds(null);
-        fitBox(bounds);
-        // Reset min zoom and movement
+        // Using padding, keep enough room for controls (zoom) to make sure they don't hide anything
+        map.fitBounds(box as LngLatBoundsLike, {
+            animate: false,
+            padding: 40,
+        });
+        // Set new map max bounds after bbox changes
         map.setMaxBounds(map.getBounds());
```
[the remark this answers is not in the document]
RafaelSzmarowski: I'm not sure I get your point here, if we don't undo the `maxBounds` here we can"t `fitBounds` properly so we can't then `getBounds` and adapt `setMaxBounds`... 

All this bounds are a bit confusing but I don't see a clear another solution here if you want to make the map automatically resize, sometimes change the map canvas ratio and still properly fit the map bounds and restrict the bounds than just do all the actions when the map canvas changes. Splitting all those actions seemed to lead to many bugs as they are not really independent and seem all useful in our case. setMaxBounds should adapt to the new bbox but also to the new size of the map canvas. So unless I'm missing something I don't see how you can achieve that separately but I'm not a Svelte expert and there might be a way to do that differently, if you have some tips here it would be great 😃

PR #148, opened 2023-01-23
New options for the select filter

KevinFabre-ods:
## Summary

The goal for this PR is to add customization to the select filter with two new props: multiple and placeholder

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-37067](https://app.shortcut.com/opendatasoft/story/37067).

### Changes

New options for the select filter:
- placeholder: Label displayed as the first select option
image
- multiple: Allow to select more than one option
image

### Changelog

Add two options for the select filter (multiple & placeholder)

## Review checklist

- [x] Description is complete

### DIFF
--- a/packages/visualizations/src/filters/components/Select/Select.svelte
+++ b/packages/visualizations/src/filters/components/Select/Select.svelte
@@ -13,6 +13,4 @@
     $: component = multiple ? MultiSelect : SingleSelect;
 </script>
 
-<div class="filter-select">
-    <svelte:component this={component} {...rest} on:filter />
-</div>
+<svelte:component this={component} {...rest} on:filter />
--- a/packages/visualizations/src/filters/components/Select/Select.svelte
+++ b/packages/visualizations/src/filters/components/Select/Select.svelte
@@ -3,7 +3,7 @@
     This is a minimal implementation of a filter, based on a standard HTML component.
     It mostly exists for demo and documentation purposes, and may be removed at a later time.
     */
-    import { exactMatch } from '@opendatasoft/api-client';
+    import { exactMatch, one } from '@opendatasoft/api-client';
     import { createEventDispatcher } from 'svelte';
     import type { DispatchedFilterEvent, SelectOptions } from '../../types';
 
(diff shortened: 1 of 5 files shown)


packages/visualizations/src/filters/components/Select/Select.svelte
```diff
@@ -3,34 +3,49 @@
...
+    function applyFilter(newValue: string | string[]) {
+        if (typeof newValue === 'string')
+            dispatch('filter', {
+                value: newValue ? exactMatch(fieldName, newValue) : null,
+            });
+        else {
+            dispatch('filter', {
+                value: one(...newValue.filter(Boolean).map((v) => exactMatch(fieldName, v))),
+            });
+        }
```
etienneburdet, line 27: I'd rather split that into two components, since each part correspond to it's own templating part anyways. 

We could still have a top component that call each subcomponent

packages/visualizations/src/filters/components/Select/Select.svelte
```diff
@@ -3,34 +3,49 @@
...
+            {/each}
+        </select>
+    {:else}
+        <select bind:value={selected}>
+            <option value>{placeholder || 'Select a value'}</option>
+            {#each availableValues as availableValue}
+                <option value={availableValue}>{availableValue}</option>
+            {/each}
+        </select>
+    {/if}
```
etienneburdet, line 48: I would do either something like:
```suggestion
    {#if multiple}
      <MultiSelect />
    {:else}
        <SingleSelect /h»
    {/if}
```

Or:
```suggestion
$: component = multiple ? MultiSelect : SingleSelect

<svelte:component this={component}>

etienneburdet (changes requested):
The queries are looking good and it works overall.

I think we should have subcomponents here , since the complexity will quickly grow and with have matching if/else in script and template. 

The multiple selection will most likely quickly grow into its own thing, but I'm for merging it like this for now. Later we could just make the MultiSelect component more complex.

KevinFabre-ods:
@etienneburdet 

It is what you had in mind f648a674e18d074b3a590854cadb9ec35f567a1b ?

etienneburdet:
Yes, something like that indeed. Looks great! 👌

wmai (changes requested):
I don't have a strong opinion a that but do we really need to have a `Select` component that makes the switch between `SingleSelect` and `MultiSelect`? I feel like we could switch between those two in the Studio instead. Maybe you had something in mind for futur implementations?

Moreover, I have a mysterious bug regarding the styling of the components, there's no styling applied, and I dont't understand why 😬 Do you experience the same issue?

KevinFabre-ods:
@wmai 

> I don't have a strong opinion a that but do we really need to have a `Select` component that makes the switch between `SingleSelect` and `MultiSelect`? I feel like we could switch between those two in the Studio instead. Maybe you had something in mind for futur implementations?

I have no opinion on this implementation as this is temporary until we develop the proper select.
BTW, With Rafael, we are more on the side to expose one component that handles all the options.

> Moreover, I have a mysterious bug regarding the styling of the components, there's no styling applied, and I dont't understand why 😬 Do you experience the same issue?

Not I don't, maybe a cache issue. @RafaelSzmarowski when you'll do the review could you please tell use if you have also no styling applied ? Thx

RafaelSzmarowski:
> 

One Select to rule them all... Yes, indeed, I think probably all the Select options will be handled through props instead, especially if we decide to implement that based on svelte-select, it seems like it will be maybe useless to have different components at start. We will probably make something more like Chart.svelte.

What do you mean @wmai by no style ? on SDK or Studio ? 
I have this in my Storybook:
Capture d’écran 2023-01-27 à 15 10 19

RafaelSzmarowski (commented):
Good addon to the filters ! 

The queries are looking well and the placeholder seems to work as expected.

At this point, I don't think that we the need the Single and Multi split as the code looks really the same for both. The only big "difference" is in the `applyFilter` function so far but I don't think that we really need a Svelte component to compute each `applyFilter` function, but maybe I'm not fully aware of the Svelte standards.

The multiple select works as expected but really feels broken as it is always open and the selection is not made by clicking but by a dragged mouse left keydown.

I have the feeling that it's maybe time to try a svelte-select implementation with only one Select.svelte (like the Chart.svelte model), isolate the `applyFilter` complexity in a separate file if needed (but it can be made later I think). 

Do you want to try it here or in some other branch ?

KevinFabre-ods:
> 



> Good addon to the filters !
> 
> The queries are looking well and the placeholder seems to work as expected.
> 
> At this point, I don't think that we the need the Single and Multi split as the code looks really the same for both. The only big "difference" is in the `applyFilter` function so far but I don't think that we really need a Svelte component to compute each `applyFilter` function, but maybe I'm not fully aware of the Svelte standards.
> 
> The multiple select works as expected but really feels broken as it is always open and the selection is not made by clicking but by a dragged mouse left keydown.
> 
> I have the feeling that it's maybe time to try a svelte-select implementation with only one Select.svelte (like the Chart.svelte model), isolate the `applyFilter` complexity in a separate file if needed (but it can be made later I think).
> 
> Do you want to try it here or in some other branch ?

I agree with your observation.
Let's try in this branch 😃

etienneburdet:
I didn't look in details, but the component split was because of there was matching templating/logic in matching ifs. If `svelte-select` removes that, it might be not be ideal indeed.

KevinFabre-ods:
As discussed, we will replace filters components with ODSQL query utilities.

PR #149, opened 2023-01-24
Chore/immutability

etienneburdet:
## Summary

The goal for this PR is to enforce immutability on all Svelte component at compiler level. Since that what we want, we might as well enforce it as default rather than execption

It is still possible to opt out with `<svelte:options immutable={false} />`
### Changes
Modified mainly the Chart component that using a big object with updates for storing states. I decided to use the same `immutability-helper` that we use in the API client. It's just a convenience over `Object.assing({}, …)`… but convenient nonetheless 😅 We could use Ramda or Lodash, I don't mind really.

For other components it was mainly a check and removing `<svelte:options immutable />` that is not necessary anymore.

## Open discussion
Arguably we would also want to remove patterns like:
```
let reactive;

// drop
function() { const stuff = reactive *2; …}

//rather
function(reactiveVar) { const stuff = reactiveVar * 2; … }
```
in favor of adding an extra parameter to the function and using `reactive` as an attribute when called. 

This would remove some unsafe case (if the data actually mutates, or in rare case function unsafe equality checks). This would also had the added benefit of hoisting all those functions outside of the component scope and avoid some rerenders.

I left it out though, since it's quite a big overhaul and I did not spot code like this that would mess with immutability (functions are rerceated every props change in what I see).

### DIFF
--- a/packages/visualizations/package-lock.json
+++ b/packages/visualizations/package-lock.json
@@ -18,6 +18,7 @@
                 "chartjs-plugin-stacked100": "^1.2.0",
                 "chroma-js": "^2.1.2",
                 "d3-geo": "^3.0.1",
+                "immutability-helper": "^3.1.1",
                 "lodash": "^4.17.21",
                 "luxon": "^2.0.2",
                 "maplibre-gl": "2.1.9",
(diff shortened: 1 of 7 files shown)


wmai (approved):
Thank @etienneburdet 🙏

richterb (approved):
Thanks!

PR #154, opened 2023-03-16
Choropleth: Use case-insensitive matching for shape filtering

richterb:
## Summary

The goal for this PR is to use case-insensitive matching in Choropleth's shape filtering.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-39521]

### Changes
There are some counties (UK) where the administrative boundaries from our reference data contains uppercase alphanum characters (`E08000003`), but due to various technical reasons, it ends up passed as `e08000003`, so the matching doesn't work, and we filter everything out including the one thing we didn't want.
Considering solving the issue on the other end (platform) isn't realistic without adding an additional API call, and that there is 0 chance in real life that a shape identifier will mean different things with a different case, it seems reasonable to transform the matching to be case insensitive.

Adding `["get", key]` in the expression also required passing an explicit `literal` expression for the values, since before it was tolerated to just list the values inline.

#### Breaking Changes
None in theory

## Review checklist

- [ ] Description is complete

### DIFF
--- a/packages/visualizations/src/components/Map/WebGl/ChoroplethVectorTiles.svelte
+++ b/packages/visualizations/src/components/Map/WebGl/ChoroplethVectorTiles.svelte
@@ -40,7 +40,7 @@
     let legend: MapLegend | undefined;
     let attribution: string | undefined;
     let filter: MapFilter | undefined;
-    let filterExpression: (string | string[])[] | undefined;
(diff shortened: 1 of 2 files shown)


RafaelSzmarowski (approved):
LGTM ! 🚀

PR #155, opened 2023-03-28
Upgrade Svelte to 3.57.0

richterb:
## Summary

The goal for this PR is to upgrade Svelte and its direct dependencies to the latest version. The main reason behind this is to integrate the support of `aria-description`: https://github.com/sveltejs/svelte/blob/master/CHANGELOG.md#3465

### Changes

Just upgrades to Svelte and other svelte core packages. I didn't upgrade all the other packages like the linter, because we may want to also bump TS and other stuff at the same time.

#### Breaking Changes

_describe the changes that are not backward compatible_

### Changelog

## Open discussion

## To be tested

## Review checklist

- [ ] Description is complete

### DIFF
--- a/packages/visualizations/package-lock.json
+++ b/packages/visualizations/package-lock.json
@@ -10,7 +10,6 @@
             "license": "MIT",
             "dependencies": {
                 "@mapbox/geo-viewport": "^0.5.0",
-                "@opendatasoft/api-client": "^0.6.1",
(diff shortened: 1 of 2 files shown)


richterb:
Turns out the `aria-description` stuff is just for linting, and not for another issue that wasn't related. We'll do blanket updates of all packages at another time.

PR #156, opened 2023-03-28
Add description to Chart and Map, remove ariaLabel from Chart

richterb:
## Summary

The goal for this PR is to introduce the support of alternative description of maps and charts, for accessibility purposes.

### Changes
The description itself is implemented differently for both cases:
- For Charts, I followed [Chart.JS examples](https://www.chartjs.org/docs/latest/general/accessibility.html)
- For Maps, I used `aria-description` which seems to fit the situation

I also removed `aria-label` from Charts, because I think it is redundant now that we have both a title and a description.

### Changelog

## Open discussion

## To be tested

## Review checklist

- [ ] Description is complete


packages/visualizations-react/stories/Chart/StudioLayouts/AreaChart.stories.tsx
```diff
@@ -28,7 +28,6 @@ const AreaTitleAxisGridDotsArgs: Props<DataFrame, ChartOptions> = {
     options: {
         labelColumn: 'x',
         source: defaultSource,
-        ariaLabel: 'Area chart with title, axis, grid and dots',
```
KevinFabre-ods, line 31: As `StudioChoropleth.stories.tsx` has a description in `options`. 
Maybe we should add `description` in all Chart studio story too.

KevinFabre-ods (approved):
👌

packages/visualizations-react/stories/Chart/StudioLayouts/AreaChart.stories.tsx
richterb: I added it in only one of each, just to make sure they are hidden (Chromatic would tell us if a story suddenly had text appearing). I don't think we need to add it everywhere, just that we have at least one of each case (with and without)

PR #162, opened 2023-05-15
Chart HTML Legend

etienneburdet:
## Summary

The goal for this PR is to add HTML category legend to pie charts (and basis for others) that can be spaced properly. 

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-40435](https://app.shortcut.com/opendatasoft/story/40435).

### Changes
Although it doesn't make for the most minimal PR, I changed the markup quite a bit for the following reasons:
* Changed figure and div so that figure only has one figcaption.
* The figcaption and the canvas are now the only elements gooing vertical/horizontal.
* Changed layout to flex since I didn't any other possible layout than vertical for now.
* Moved the category legend to auto-grid for having both alignment and auto-wrap without having to distinguish alignment. With flex it was necessary to specify a flex basis and not practical to center. 
* Tried to remove as many class prefix as possible, since they aren't required with scoping.

Some of these change should break snapshots a bit in the Studio (markup name change), but ideally shouldn't change Chromatic snapshots.

#### Breaking Changes

_describe the changes that are not backward compatible_

### Changelog

## Open discussion

## To be tested

## Review checklist

- [ ] Description is complete


packages/visualizations/src/components/Chart/legend.ts
```diff
@@ -67,3 +67,67 @@ export default function buildLegend(
...
+    return {
+        type: 'category' as const,
+        position: defaultValue(options?.legend?.position, 'bottom'),
+        items: chartConfig.data.datasets[0].data.map((_data, i) => ({
+            color: series[0].backgroundColor?.[i],
+            borderDashed: false,
+            label: buildLegendLabels(i, options, chartConfig),
+            onClick: (index: number) => {
+                if (chart) {
+                    chart.toggleDataVisibility(index);
```
RafaelSzmarowski, line 102: Using `toggleDataVisibility` here causes a crash in `pieDataLabelsPlugin.ts` as we do not check there if data is visible and we try to draw lines on hidden data : 

On line 22 in `pieDataLabelsPlugin.ts` I think we should wrap the drawing inside a check 
`if (chart.getDataVisibility(i)) { ... }`
This way there is no crash and the hover seems to work again !

packages/visualizations/src/components/Chart/Chart.svelte
```diff
@@ -69,14 +73,15 @@
     $: {
         // Reactively update chart configuration
         const chartOptions = chartConfig.options || {};
+        const legend = options.legend?.custom ? { display: false } : buildLegend(options);
```
RafaelSzmarowski, line 76: Just a detail, but maybe a small comment here or a FIXME as it's not easy to understand why there is a `display: false` and why we are doing the split of the legend this way ?

packages/visualizations/src/components/Chart/Chart.svelte
```diff
@@ -185,29 +198,47 @@
...
         width: 100%;
         flex-grow: 1;
+        margin: 0;
+    }
+    .legend--bottom {
+        flex-direction: column;
+    }
+    .legend--right {
+        flex-direction: row;
+    }
```
RafaelSzmarowski, line 235: It doesn't seem to work for me, I have a message in the inspect tools saying : `The display: block property prevents flex-direction from having an effect.` and the legend position never changes whatever option I pass. It seems that there is a default `display: block` on the figure html element.

packages/visualizations/src/components/Chart/Chart.svelte
```diff
@@ -156,14 +161,22 @@
...
+    $: if (options?.legend?.custom) {
+        legendOptions = buildCustomLegend({ chart, options, chartConfig });
+    }
 </script>
 
 {#if data.error}
     Error : {JSON.stringify(data.error)}
 {:else if options}
-    <figure>
+    <div bind:clientWidth class="container">
```
RafaelSzmarowski, line 177: I'm just wondering if is there a way to compute clientWidth and bind it conditionally as for now we only use it for custom legend but we process it for all charts ?

packages/visualizations/src/components/Chart/Chart.svelte
```diff
@@ -185,29 +198,47 @@
...
-        </div>
+            {#if options?.legend?.custom}
+                <figcaption>
+                    <CategoryLegend {legendOptions} />
+                </figcaption>
+            {/if}
+        </figure>
         {#if options.source}
             <SourceLink source={options.source} />
         {/if}
```
RafaelSzmarowski, line 209: The legend and view source looks a bit too close IMO, what do you think ? 
Capture d’écran 2023-05-22 à 17 36 08

packages/visualizations-react/stories/Chart/StudioLayouts/PieChart.stories.tsx
```diff
@@ -140,3 +140,48 @@ const PieTitleLegendValuesArgs: Props<DataFrame, ChartOptions> = {
...
+                    display: true,
+                },
+            },
+        ],
+        padding: 32,
+        legend: {
+            display: false,
+            custom: true,
+            position: 'right',
+            align: 'center',
```
RafaelSzmarowski, line 173: If I'm not mistaken, I haven't seen the option align used in the custom legend ?

packages/visualizations/src/components/Legend/types.ts
```diff
@@ -0,0 +1,42 @@
+import type { Color } from '../types';
+
+export const LEGEND_POSITIONS = {
+    bottom: 'bottom',
+    top: 'top',
+    left: 'left',
+    right: 'right',
+} as const;
+export type LegendPositions = typeof LEGEND_POSITIONS[keyof typeof LEGEND_POSITIONS];
```
RafaelSzmarowski, line 9: I don't know how much we want to be precise in our types, now when we use custom legend we are able to write `left` and `top` in position but it has no effect on our legend as `legend-left` and `legend-top` classnames are not defined. It depends on the future and I don't know if we should restrict it now or leave it as it is with a small comment maybe ?

RafaelSzmarowski (changes requested):
Well done for this custom legend revival ! I've seen some points that need some attention (in particular the crash in our custom pie lines plugin when using toggleDataVisibility) and left few comments but overall it works as expected 👏

packages/visualizations/src/components/Legend/CategoryLegend.svelte
```diff
@@ -0,0 +1,91 @@
+<svelte:options immutable={false} />
```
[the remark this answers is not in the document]
etienneburdet: Ha yes true, I think I tried to turn it on and off or wanted to make true the default and have this one be the exception 🤔 Whatever, not needed.

packages/visualizations/src/components/Legend/types.ts
etienneburdet: I don't know what we want indeed, but with different types imports from ChartJS (and maybe Maplibre) it just seemed easier and not that risky to accept all at this point.

etienneburdet:
I added the padding, a comment for future note about not displaying chartJS legend, removed `top` in positions, but left `left` since it used by maps.

packages/visualizations/src/components/Legend/CategoryLegend.svelte
```diff
@@ -0,0 +1,95 @@
...
+        on:mouseleave={() => {
+            if (item.onLeave) {
+                item.onLeave();
+            }
+        }}
+    >
+        {#if item.color}
+            <div
+                class="color-box-category"
+                class:refined={refinedSeries[i]}
```
RafaelSzmarowski, line 38:
```suggestion
                class:refined={Boolean(refinedSeries.find((color) => color === item.color))}
```
It doesn't seem safe to use index here, while testing it appears that it can easily not match the refined array (tested in storybook)

RafaelSzmarowski: Capture d’écran 2023-05-31 à 17 38 30

To reproduce the unwanted behaviour, if you click first on the blue box then it disappears from the chart but the red box is marked as refined

etienneburdet: Ha yes, left over from the previous method 👌

packages/visualizations/src/components/Legend/CategoryLegend.svelte
```diff
@@ -0,0 +1,94 @@
...
+    import type { CategoryLegend, CategoryItem } from './types';
+
+    export let legendOptions: CategoryLegend;
+    let refinedSeries: string[] = [];
+
+    const isRefined = (item: CategoryItem, series: string[]) =>
+        series.some((color) => item.color === color);
+</script>
+
+{#each legendOptions.items as item, i (item.color)}
```
RafaelSzmarowski, line 11: I don't know if this error is really due to this part of code but it seems that we can't have twice the same colour for series... 

Capture d’écran 2023-06-01 à 09 53 22

It seems also logical to not use colour as the identifier to refine series to avoid refining multiple series in one click... Probably we should give them unique ids

packages/visualizations/src/components/Legend/CategoryLegend.svelte
```diff
@@ -0,0 +1,97 @@
...
+    const isRefined = (item: UniqueCategoryItem, series: number[]) =>
+        series.some((id) => id === item.id);
+</script>
+
+{#each categoryItems as item (item.id)}
+    <div
+        class="color-item-category"
+        on:click={() => {
+            // using color as a unique id proxy
+            if (item?.color) {
```
RafaelSzmarowski, line 19: I guess this check on `color` and comment are a bit obsolete now

RafaelSzmarowski (approved):
LGTM ! I've left just a minor remark on an obsolete comment

richterb (approved):

PR #166, opened 2023-05-25
Changes to the api-client to use the api v2.1

GaetanMartinODS:
## Summary

The goal for this PR is to switch to v2.1 of the API.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-40301](https://app.shortcut.com/opendatasoft/story/40301/sdk-api-client-update-to-explore-api-2-1?ct_workflow=all&cf_workflow=2).

### Changes

- change : changes to search function, it now accept an optional text_fields param to specify the fields to search in.
- add : textStartWith & textSuggest, search function that use new API v2.1 function
- add enum to enforce correct export format

#### Breaking Changes

- change : remove the query() function and related query endpoint
- change : remove aggregates() function, was already deprecated
- XLS export format is removed and changed to XSLX

### Changelog

Going out of beta, the api-client is now using the version2.1 of the API. 

## Review checklist

- [x] Description is complete


packages/api-client/README.md
```diff
@@ -1,6 +1,6 @@
 # @opendatasoft/api-client ![CI status](https://github.com/opendatasoft/ods-dataviz-sdk/workflows/CI/badge.svg)
 
-This package implements a Typescript/Javascript client library for [Opendatasoft's Search APIv2](https://help.opendatasoft.com/apis/ods-search-v2/#search-api-v2).
+This package implements a Typescript/Javascript client library for [Opendatasoft's Search APIv2.1](https://help.opendatasoft.com/apis/ods-explore-v2/explore_v2.1.html).
```
KevinFabre-ods, line 3: The `README.md` at the root of the project has also a link to the explore api documentation

packages/api-client/src/client/types.ts
```diff
@@ -66,5 +66,30 @@ export interface ApiFacets {
...
+
+export enum ExportCatalogFormat {
+    CSV = 'csv',
+    JSON = 'json',
+    XLSX = 'xlsx',
+    RDF = 'rdf',
+    TTL = 'ttl',
+    DATA_SJON = 'data.json',
+    RSS = 'rss',
+    DCAT = 'dcat',
```
KevinFabre-ods, line 80: Is there a reason to not list `dcat_ap_*` format ?

packages/api-client/src/odsql/clauses.ts
```diff
@@ -1,11 +1,36 @@
...
 
+const formatFields = (fields: string[]) => {
+    if (!fields || fields.length === 0) return '';
+    return `${fields.map(field).join(', ')},`;
+};
+
+/**
+ * Perform full-text search on all fields. It matches the text fields that contain strings beginning with the searched string.
+ * @param text_fields Fields to search in
+ * @param text Text to search for
```
KevinFabre-ods, line 11:
```suggestion
 * @param text Text to search for
 * @param text_fields Fields to search in
```

To follow args order

Same for `textStartWith` and `textSuggest`

packages/api-client/src/odsql/clauses.ts
```diff
@@ -1,11 +1,36 @@
 import { field, string } from '.';
 
+const formatFields = (fields: string[]) => {
+    if (!fields || fields.length === 0) return '';
+    return `${fields.map(field).join(', ')},`;
+};
+
+/**
+ * Perform full-text search on all fields. It matches the text fields that contain strings beginning with the searched string.
+ * @param text_fields Fields to search in
```
KevinFabre-ods, line 10: `text_fields` looks optional ?

I think the description for this arg should indicate that the search will be restricted to the fields passed.

packages/api-client/src/odsql/clauses.ts
```diff
@@ -1,11 +1,36 @@
 import { field, string } from '.';
 
+const formatFields = (fields: string[]) => {
+    if (!fields || fields.length === 0) return '';
+    return `${fields.map(field).join(', ')},`;
+};
+
+/**
+ * Perform full-text search on all fields. It matches the text fields that contain strings beginning with the searched string.
```
KevinFabre-ods, line 9: Looks like the description for the `textStartWith` function.

Also apply for the `textSuggest` function

packages/api-client/src/odsql/clauses.ts
KevinFabre-ods, line 1: It could be nice to start testing this clauses 😎

packages/api-client/src/odsql/clauses.ts
```diff
@@ -1,11 +1,36 @@
...
+ * @param text_fields Fields to search in
+ * @param text Text to search for
+ * @returns An ODSQL query, ready to be used as a `where` clause
+ */
+export const textSearch = (text: string, text_fields: string[] = []) =>
```
KevinFabre-ods, line 14: Is there a reason to use snake case instead of camel case ?

KevinFabre-ods (changes requested):
Nice work, 

I think some changes could improve the readableness of the clauses functions.

KevinFabre-ods:
Can you also follow the contribution guidelines https://github.com/opendatasoft/ods-dataviz-sdk/blob/main/CONTRIBUTING.md#commit-messages ? 

Good commit messages help our release process

packages/api-client/src/client/index.ts
```diff
@@ -48,7 +48,7 @@ function computeBaseUrl(domain: string): string {
     if (!baseUrl.endsWith('/')) {
         baseUrl += '/';
     }
-    baseUrl += 'api/v2/';
+    baseUrl += 'api/explore/v2.1/';
```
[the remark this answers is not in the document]
GaetanMartinODS: I dunno the reason, I think they just depreciate the old path without /explore/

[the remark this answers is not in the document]
richterb: Yes, this is for clarity purposes (we have all our APIs below `/api/` such as `/api/automation/` etc, Explore was the only one that didn't have a root before.

packages/api-client/src/odsql/clauses.ts
GaetanMartinODS: Absolutly no reason. I read too much python these days.

packages/api-client/src/client/types.ts
richterb: I think it was added just a few days ago, now it appears in the doc

packages/api-client/src/client/types.ts
```diff
@@ -66,5 +66,30 @@ export interface ApiFacets {
...
+    records: T[];
+}
+
+export enum ExportCatalogFormat {
+    CSV = 'csv',
+    JSON = 'json',
+    XLSX = 'xlsx',
+    RDF = 'rdf',
+    TTL = 'ttl',
+    DATA_SJON = 'data.json',
```
richterb, line 78: Tiny typo SJON/JSON 🙈

packages/api-client/src/client/types.ts
GaetanMartinODS: Oh ok. Do you know if more format are gonna be added often ? Maybe a fixed enum isn't the best idea if format are gonna be added quite often.

richterb: It was in 2.1 from the start, just not in the doc, it probably was a mistake. These formats are a bit special because they are activated per-domain, so they are not available everywhere (but the API Client has no way to know that), this is probably why the doc didn't mention them at first. But yes we could double-check if formats can be added in a stable version (after all, it would be a non-breaking change), maybe we shouldn't use an enum indeed

packages/api-client/src/client/types.ts
```diff
@@ -66,5 +66,36 @@ export interface ApiFacets {
 
 export interface ApiQuery<T> {
     total_count?: number;
-    results: T[];
+    records: T[];
+}
+
+export enum ExportCatalogFormat {
```
etienneburdet, line 72: I'm more and more using: 
```
ExportCatalogFormat = {
    ...
} as const;
ExportCalalogFormat type = typeof ExportCatalogFormat[keyof typeof ExportCatalogFormat];
```
It seems a bit verbose, but objects are useable in vanilla JS (or enums are exported as objects maybe?) and more often than not, you end up building the object and according type anyways.

etienneburdet: (got that from the [TS doc](https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums) by the way)

etienneburdet (approved):
Seems to work great on my side 👌  Slight remark on Enum, but more of style choice than anything.

If you want to keep pushing on these topic, manu mentionned adding tests against real API, which is something we don't really know how to do yet…

packages/api-client/src/client/types.ts
richterb: I got an error when building the package and I can't see why it would say that considering the code:
```
@opendatasoft/api-client: (!) Plugin typescript: @rollup/plugin-typescript TS2345: Argument of type '"geojson"' is not assignable to parameter of type 'ExportDatasetFormat'.
```
I get issues with TS and enums so many times, I'm starting to think enums are more trouble than they're worth in some cases.

If we got too many issues we could also use a good one `ExportCatalogFormat = 'json' | ' csv';` etc etc.

But maybe Etienne's suggestion fixes this already

packages/api-client/src/odsql/index.ts
```diff
@@ -134,24 +135,14 @@ export class Query {
...
         itself: () => new Query(`${source}/`),
-        /**
-         * @deprecated Use query() instead
-         */
-        aggregates: () => new Query(`${source}/aggregates/`),
         facets: () => new Query(`${source}/facets/`),
         datasets: () => new Query(`${source}/datasets/`),
-        query: () => new Query(`${source}/query/`),
-        exports: (format: string) => new Query(`${source}/exports/${format}/`),
+        exports: (format: ExportCatalogFormat) => new Query(`${source}/exports/${format}/`),
```
richterb, line 140: This could be outside the scope of this PR and is kind of eye-candy, but it would be a breaking change so it may be a good time. For datasets, we have:
- `.datasets()` matching `/datasets/`
- `.dataset('xxx')` matching `/datasets/xxx/`

Maybe for exports it would make more sense and be more intuitive to have `.export('geojson')` (singular) matching `.../exports/geojson/`. What do you think?

richterb (commented):
Thanks for the huge work, it feels nice to see a PR with a version that doesn't start with a `0.` 😄 

I have a few remarks, mostly doc, and some things that we already there before your PR but could be adressed before the first stable release.

packages/api-client/src/odsql/clauses.ts
```diff
@@ -1,11 +1,36 @@
...
+ * @param text Text to search for
+ * @param textFields Fields to search in
+ * @returns An ODSQL query, ready to be used as a `where` clause
+ */
+export const textStartWith = (text: string, textFields: string[] = []) =>
+    `startswith(${formatFields(textFields)}"${text}")`;
+
 /**
- * Perform full-text search on all fields. It matches the text fields that contain strings beginning with the searched string.
+ *  Perform full-text search on fields, either ones specified in textFields, or all fields. It matches the text fields that contain the searched string. Equivaqlent to Search.
```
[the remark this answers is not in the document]
KevinFabre-ods: There is no difference in the description between the suggest and search functions.
We should explain what they are about :
- search is no a fuzzy search
- suggest is now the old search

KevinFabre-ods (changes requested):
Descriptions for clauses should be updated.
As an ODS dev, it required some time to understand the changes between 2.0 and 2.1

Currently, for an external dev, IMO, the description are confusing

KevinFabre-ods:
@GaetanMartinODS, could you also resolve conflicts with this branch :blush:

packages/api-client/src/client/types.ts
GaetanMartinODS: I didn't know this const TS trick.  I'm trying it

packages/api-client/src/odsql/clauses.ts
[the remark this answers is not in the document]
GaetanMartinODS: You're right, I totally misread the equivalent part. Correcting it.

packages/api-client/src/odsql/index.ts
GaetanMartinODS: Hmm I think you're right, this is the best moment to do this.

KevinFabre-ods (changes requested):
Sorry for the nitpicking, but I think a good description is essential to help external developers understand how the search methods differ from each other.

KevinFabre-ods (approved):
Thanks for the extra miles 🚀

richterb (approved):
I just realized I had pending comments I didn't send, but Kevin's review pointed out mostly the same things, so what's left is keeping "Explore API" as a brand. Everything else looks good to me and we're ready for prime time!

We'll have to be careful during the first release so that Lerna doesn't increment the version number or something and we push 21.0.1 on NPM (@wmai our lerna publish jedi master 👀 )

README.md
```diff
@@ -2,7 +2,7 @@
 
 A monorepo for the Opendatasoft's libraries:
 
-- [`@opendatasoft/api-client`](packages/api-client/): An API client for the [Opendatasoft's Explore APIv2](https://help.opendatasoft.com/apis/ods-explore-v2/).
+- [`@opendatasoft/api-client`](packages/api-client/): An API client for the [Opendatasoft's APIv2.1](https://help.opendatasoft.com/apis/ods-explore-v2/explore_v2.1.html).
```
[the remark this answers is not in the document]
richterb: fixed in ed9c27064bb04c6c7f6d300ec8ad9230b3c578bd

wmai:
> We'll have to be careful during the first release so that Lerna doesn't increment the version number or something and we push 21.0.1 on NPM (@wmai our lerna publish jedi master 👀 )

Screenshot 2023-06-14 at 14 20 10

Good catch @richterb. I suppose we'll need to choose "`Custom Version`" and manually type in `21.0.0` to avoid this issue ^^

richterb:
After our second look based on the real outputs from 2.1, I made the following changes: 668caf66b945c937de4acb4d18249b9f9232acf8

Some types were rendered useless because they returned directly the structure of a dataset / record now (e.g. `ApiRecord<T> is effectively `OdsRecord<T>` now).

packages/api-client/README.md
```diff
@@ -57,7 +57,7 @@ client
     .catch(error => console.error(error));
 ```
 
-[CodeSandbox sample](https://codesandbox.io/s/api-clientget-started-be0xu?file=/src/index.js)
+[CodeSandbox sample](https://codesandbox.io/s/api-client-get-started-v2-1-4jpvd9)
```
wmai, line 60: In the sand box, we try to access the key `results` in the promise resolution. But we should access `records` instead.

PR #167, opened 2023-05-30
Feature/sc 40297/sdk api client support api deprecation header

GaetanMartinODS:
## Summary

Follow up of this PR : https://github.com/opendatasoft/ods-dataviz-sdk/pull/166#event-9378374403

The goal for this PR is to show a warning when using a deprecated api function

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-40297](https://app.shortcut.com/opendatasoft/story/40297/sdk-api-client-support-api-deprecation-header).

### Changes

Add a warning when using a deprecated API function
Add an option to disable those warning (so that when using the API in production, you can disable it)

### Changelog

When using a deprecated API function,  warning appear in thje console. An option is avaible to disable those warning

## Review checklist

- [ ] Description is complete

### DIFF
--- a/packages/api-client/src/client/index.ts
+++ b/packages/api-client/src/client/index.ts
@@ -86,6 +86,8 @@ function buildConfig(
 export class ApiClient {
     readonly defaultConfig: ApiClientConfiguration;
 
+    deprecatedWarningShown: string[];
+
     /**
      * Constructs an instance of {@link ApiClient}
      */
@@ -136,7 +138,14 @@ export class ApiClient {
         // Send request
         const { fetch } = config;
         const fetchResponse = await fetch(request);
-        if (!options?.hideDeprecatedWarning) console.warn(`@opendatasoft/api-client : ${fetchResponse.headers.get('Ods-Explore-Api-Deprecation')}`);
+        if (!options?.hideDeprecatedWarning) {
+            const msg = fetchResponse.headers.get('Ods-Explore-Api-Deprecation');
+            if(msg !== null && !this.deprecatedWarningShown.includes(msg)) {
+                this.deprecatedWarningShown.push(msg);
+                // eslint-disable-next-line no-console
+                console.warn(`@opendatasoft/api-client : ${msg}`);
+            }
+        } 
         if (config.interceptResponse) return config.interceptResponse(fetchResponse);
 
         if (fetchResponse.ok) {
--- a/packages/api-client/src/client/index.ts
+++ b/packages/api-client/src/client/index.ts
@@ -13,10 +13,10 @@ const _global: any =
     typeof global !== 'undefined'
         ? global
         : typeof self !== 'undefined'
-        ? self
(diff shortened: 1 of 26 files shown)


etienneburdet (approved):
It's working 😛 !

packages/api-client/src/client/index.ts
```diff
@@ -127,6 +136,7 @@ export class ApiClient {
         // Send request
         const { fetch } = config;
         const fetchResponse = await fetch(request);
+        if (!options?.hideDeprecatedWarning) console.warn(`@opendatasoft/api-client : ${fetchResponse.headers.get('Ods-Explore-Api-Deprecation')}`);
```
richterb, line 139: Shouldn't it only appear if there is a content in the header? Right now it runs even if the header is null.

Also, shouldn't we only run this once, to avoid spamming the console? We could store a `deprecationWarned` internal boolean var in the ApiClient class for that. In effect it would run once per instantation of the client, so once per page, once per run of a script etc.

packages/api-client/src/client/index.ts
```diff
@@ -70,6 +74,11 @@ function buildConfig(
     if (fetch) newConfig.fetch = fetch;
     if (interceptRequest) newConfig.interceptRequest = interceptRequest;
     if (interceptResponse) newConfig.interceptResponse = interceptResponse;
+    if (hideDeprecatedWarning) {
+        newConfig.hideDeprecatedWarning = hideDeprecatedWarning;
+    } else {
+        newConfig.hideDeprecatedWarning = false;
+    }
```
richterb, line 81: Since it's a boolean, we could keep the same runtime safety with this I think
```suggestion
        newConfig.hideDeprecatedWarning = Boolean(hideDeprecatedWarning);
```

richterb (changes requested):
We should mention `hideDeprecatedWarning` in the README (Usage > ApiClient) next to the other parameters.

You can change the base branch of the PR to `https://github.com/opendatasoft/ods-dataviz-sdk/tree/chore/sc-40324/integrate-api-client-21-0-0-in-platform` for a cleaner diff if you want, we usually do that for nested branches

wmai (changes requested):

packages/api-client/src/client/index.ts
richterb: Ah yes, it's better indeed!

richterb (approved):
I'm not sure how to test it "for real" (maybe there's some kind of local REPL that can open in a browser for dev, like a Storybook for non-visual stuff, but I don't know about one), but we'll do that when we test integration in the Studio. LGTM!

richterb:
As discussed offline after testing, for now the feature doesn't work cross-domain because Explore API lacks a `Access-Control-Expose-Headers` header, this will be addressed outside this PR (sc-41250). 

@wmai do you want to have a last look or do we go ahead?

PR #171, opened 2023-06-29
Feature: poi maps add tooltip

RafaelSzmarowski:
## Summary

The goal for this PR is to add a first implementation of a tooltip.

The tooltip supports keyboard navigation, we should probably add some description to help users to understand keyboard shortcuts. But I have the feeling that this should be inside an invisible map description passed as an option (to adapt to map config and language).

To navigate among tooltips: focus on map, press enter, the first feature is selected, tab navigation inside tooltip, arrow left and right to change selected feature, space or escape to close tooltip and refocus on map.

As for now, we toggle the "selected" feature state using `feature-state` property and different paint circle properties. This doesn't allow to use `circle-sort-key` to display the selected feature on top of the others if there are multiple features in the same location. Indeed, `circle-sort-key` is a layout porperty and doesn't allow to use `feature-state`. One possible workaround would be to use two layers, one for the selected features and one for the others and toggle visibility, but this seems less natural than using `feature-state` so probably we should do the second implementation only if we have strong needs at this points.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-40342](https://app.shortcut.com/opendatasoft/story/40342/sdk-show-tooltip-to-end-users).

## Review checklist

- [ ] Description is complete


KevinFabre-ods (changes requested):
Great work !

Functionally it works very well, and GG for the accessibility effort. 

I wrote a lot of comments, some are for me when I'll do the edits 😉

packages/visualizations/src/components/MapPoi/utils.ts
KevinFabre-ods, line 1: As I said [here](https://github.com/opendatasoft/ods-dataviz-sdk/pull/164/files#r1272278752), I'm not confinded about the current implementation for the `circle-color` property

packages/visualizations/src/components/MapPoi/PoiGeoJson.svelte
```diff
@@ -30,13 +32,19 @@
...
+    $: ({ fixed = false } = tooltip);
+
     let renderedBbox = bbox || VOID_BOUNDS;
 
     function computeSourceLayers(newShapes: FeatureCollection) {
         source = {
             type: 'geojson',
             data: newShapes,
+            // We need a feature id to use active feature-state to toggle selected feature
+            promoteId: 'featureId',
```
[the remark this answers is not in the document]
etienneburdet: I actually promoteId. We need an Id anyways, we might want to fetch additional data, navigate back to the feature etc. I'd rather have one of our fields as id. The doc mention using promoteId for vector and generateId for geojson.

etienneburdet (commented):
My though for now:
* Tooltip should be it's own component or package (to use with `use`), indeed. There multiple ways to have it get the map (context, prop…).
* The functions are too nested right now making it hard to know what will be recreated (`handleInteractivity` is recreated every time `handeTootip`changes, which is itself tied with multiple `let` variables. 
* The flyTo zoom should be a property (10 could be a default). That shouldn't be hard to change.

Functionnally seems great to me, except I couldn't have the keyboard navigation to work. All my keystrokes are never detected and either do stuff in my browser or move the map.

packages/visualizations/src/components/MapPoi/MapRender.svelte
```diff
@@ -45,6 +56,122 @@
...
+    function handleTooltip(
+        FeatureCollection: Feature<Geometry, GeoJsonProperties>[],
+        indexToDisplay: number
+    ) {
+        if (
+            // Check existence, ids and geometry type
+            FeatureCollection &&
+            FeatureCollection?.[indexToDisplay]?.properties?.featureId &&
+            FeatureCollection[indexToDisplay].geometry.type === 'Point'
+        ) {
```
etienneburdet, line 106: Could be a type assertion to avoid the type casting 
```suggestion
    const isPointGeometry = (geometry: Geometry): geometry is Point => FeatureCollection[indexToDisplay].geometry.type === 'Point'
        if (
            // Check existence, ids and geometry type
            FeatureCollection &&
            FeatureCollection?.[indexToDisplay]?.properties?.featureId &&
            isPointGeometry(geometry)
        ) 
```

packages/visualizations/src/components/MapPoi/MapRender.svelte
```diff
@@ -45,6 +56,122 @@
...
+        });
+    } else {
+        if (clickPopup) {
+            removeTooltip();
+        }
+        clickPopup = new maplibregl.Popup({
+            closeOnClick: false,
+            closeButton: false,
+            className: 'tooltip-on-click',
+        });
```
etienneburdet, line 94: I think we can just have a `className: fixed ? '...' : '....'`. It will recreate a newPopup if `fixed` changes.

etienneburdet (changes requested):
My main points are : 
* Tooltip should indeed be its own component or function/package. Functions like `clearActiveFeature` gets the right way to do it imo. If needed we can pass the `map` instance as a context. It would many things, including the click event.
* The functions and reactive variables starts being really nested wich makes it hard to understand what will actually be recreated (`handleInteractivity` depends on `let …` and `handleTooltip`, which itself depends on other `let`
* It feels we have a bit of double tracking of the active state, once in Svelte and once in Maplibre with the feature state. Maybe we could use `setSate` and build the full style JSON rather than `setFeatureState`? I guess it's not to optimal either to use `getFeatureState` or `setFeatureState` directly to clear the active feature.
* Sadly, I can't have the keyboard navigation to work, it always move the map or do stuff in my browser. (despite the nice effort indeed!). Probably an `aria-live-region` is the way to go (or even a list), hidden as Rafael mentions.

As for the `circe-sort-key`, another workaround could be to use a `Marker` for the active feature, rather than `feature-state`. It's extra markup, but it's flexible and we don't risk overflowing since there can only be one.

I'll let that digest a bit and try to code/discuss something.

PR #178, opened 2023-07-28
[FEATURE] PoiMap: Handle vector source

KevinFabre-ods:
**CLOSED: Clean implementation can be found [here](https://github.com/opendatasoft/ods-dataviz-sdk/pull/185)**

## Summary

The goal for this PR is to offer a new way to renders POI maps.
Instead of passing a single type of source (`geojson`) to the map, all sources can be passed.

Indeed, with ODS we are currently trying to use vector sources.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-42268](https://app.shortcut.com/opendatasoft/story/42268).

### Changes
Our usage of the POI maps requires us to be able to changes sources and layers very often. 
Sources and layers can be set in the [MapLibre style object](https://maplibre.org/maplibre-style-spec/).
So in this implementation, whenever a source or a layer changes, the full style object is rebuilt and the map updated with `map.setStyle(style)`.

Why the `data` object contain a `style` and a `sources` key ?
- `style` is for [map style configuration URL](https://maplibre.org/maplibre-style-spec/). It the only key required to render a map.
- `sources` (optional) is for all 'data sources' like vector, geojson...

Options is a light and simplify versions of all options that a MapLibre map can have.
- `bbox`: to set the bounding box
- `aspectRatio`: to set map dimension
- `interactive`: is the map can be zoom, dragged, etc...
- `layers`: simple object to style a data sources. Has its own abstraction compare to the[ Maplibre layer config](https://maplibre.org/maplibre-style-spec/layers/). **Currently supports only the `circle` layer**.

#### Technical changes

A Map class has been created to move all function related to MapLibre in a isolated file. Making the `MapRender` component cleaner. 

### Breaking Changes

none
### Changelog

none

## Open discussion

I was unable to remove the `Async` type surrounding the data. There are type checks when React components are built from SDK components.
I think we're a bit early in the project to consider changing this.

## To be tested

Currently being tested with ODS Studio

## Review checklist

- [x] Description is complete


packages/visualizations/src/components/MapPoi/types.ts
```diff
@@ -1,38 +1,50 @@
...
+/**
+ * Base on Maplibre layers https://maplibre.org/maplibre-style-spec/layers/
+ * Use only part of the configuration to build layers with consistent styles.
+ */
+export type Layer = {
+    source: string;
+    sourceLayer?: string;
+    type: LayerSpecification['type'];
+    color: Color;
+    groupBy?: {
```
KevinFabre-ods, line 43: Change name ; may suggest clustering.

packages/visualizations/src/components/MapPoi/types.ts
```diff
@@ -1,38 +1,50 @@
...
-};
+// Supported layers
+export type LayerSpecification = CircleLayerSpecification;
 
-export type PoiMapLayer = CircleLayer;
+/**
+ * Base on Maplibre layers https://maplibre.org/maplibre-style-spec/layers/
+ * Use only part of the configuration to build layers with consistent styles.
+ */
+export type Layer = {
```
KevinFabre-ods, line 38: add an `id`

PR #179, opened 2023-08-03
fix: display zero value as 0 instead of null

AnthoPepin:
## Summary

The goal for this PR is to fix a bug where 0 value was displayed as null on a percentaged bar graph

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-41137](https://app.shortcut.com/opendatasoft/story/41137).

### Changes

Instead of checking `raw?` we check that `raw!==undefined` which should be better for 0 values.

### Changelog

We fixed a bug where null was displayed instead of 0 on percentage charts

## Review checklist

- [x] Description is complete

### DIFF
--- a/packages/visualizations/src/components/Chart/Chart.svelte
+++ b/packages/visualizations/src/components/Chart/Chart.svelte
@@ -111,7 +111,7 @@
                         const percentaged = options?.axis?.assemblage?.percentaged ? '% ' : '';
                         // If the value is a percentage, we need to format the raw value
                         const formattedRawValue =
-                            percentaged && raw && typeof raw === 'number' && `(${format(raw)})`;
+                            percentaged && raw !== undefined && typeof raw === 'number' && `(${format(raw)})`;
                         const suffix = percentaged + formattedRawValue;
 
                         if (seriesType && parsed) {


packages/visualizations/src/components/Chart/Chart.svelte
```diff
@@ -111,7 +111,10 @@
                         const percentaged = options?.axis?.assemblage?.percentaged ? '% ' : '';
                         // If the value is a percentage, we need to format the raw value
                         const formattedRawValue =
-                            percentaged && raw && typeof raw === 'number' && `(${format(raw)})`;
+                            percentaged &&
+                            raw !== undefined &&
+                            typeof raw === 'number' &&
+                            `(${format(raw)})`;
```
wmai, line 117: Here, we could remove the explicit `raw !== undefined` check since typeof `raw === 'number'` already ensures that `raw` is defined and is a number. If raw was `undefined`, the type check would return `undefined`, and the expression would evaluate as `false`.

If you want to make sure `raw` is not only a number but also not `NaN`, you can add a check for that:

``` JS
const formattedRawValue = percentage && typeof raw === 'number' && !isNaN(raw) ? `(${raw})` : `(0)`;
```

wmai (changes requested):

packages/visualizations/src/components/Chart/Chart.svelte
AnthoPepin: Shoudn't `isNaN(0) === true` ?

wmai: Yes indeed

packages/visualizations/src/components/Chart/Chart.svelte
```diff
@@ -112,8 +112,8 @@
                         // If the value is a percentage, we need to format the raw value
                         const formattedRawValue =
                             percentaged &&
-                            raw !== undefined &&
                             typeof raw === 'number' &&
+                            !Number.isNaN(raw) &&
```
wmai, line 116: I might be missing something, but don't we want to return `(0)` if the raw value is `null` or `undefined`? If so, the current code won't returns the right value, it will return `false` if `raw === null` or `raw === undefined`.  

`Number.isNaN` doesn't have the same behaviour as `isNaN`: https://stackoverflow.com/a/33164924

AnthoPepin: Indeed we do want it to return `(0)` but that's what the last `(${format(raw)})` does
However, the eslint rule does not allow the use of `isNaN` but only the use of `Number.isNaN` which is quite good because we already check that `raw` is a number, we should not check it again with `isNaN`. We just need to cross the `NaN` value as it is of type `number` too

AnthoPepin: Oh right I just got your point! That's a good question, TBH I don't know what is best : display null to show that there is an error, or display 0

wmai: Sorry I didn't know about the eslint rule, `Number.isNaN` is good enough you're right.

A little bit of code my worth a thousand words ^^ Here a little codepen illustrating the issue with the current code: https://codepen.io/wmai/pen/yLQZryL

What do you think?

wmai: I'd say let's keep it simple, let's display `(0)`.

wmai (approved):

PR #180, opened 2023-08-03
feat: add-cutout-option-for-doughnut-charts

AnthoPepin:
## Summary

The goal for this PR is to allow adding a cutout to Doughnut charts as a percentage string

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-42059](https://app.shortcut.com/opendatasoft/story/42059).

### Changes

Adding a case handling where we check if `option` has a `cutout` property. We cannot use a type guard to check if we use a Doughnut chart because of a bug in ChartJS : https://github.com/chartjs/Chart.js/issues/10896#issuecomment-1660559770*/

Edit: Now we check if `chartType`is `Doughnut` so we can look for the series option `cutout` to put it in the `chartOptions`

## Review checklist

- [x] Description is complete

### DIFF
--- a/packages/visualizations-react/stories/Chart/RadialCharts/RadialCharts.stories.tsx
+++ b/packages/visualizations-react/stories/Chart/RadialCharts/RadialCharts.stories.tsx
@@ -1,0 +1,145 @@
+import type { ChartOptions, DataFrame } from '@opendatasoft/visualizations';
+import { ChartSeriesType } from '@opendatasoft/visualizations';
+import { Meta } from '@storybook/react';
+import type { Props } from '../../../src';
+import { compactNumberFormatter, defaultSource } from '../../utils';
+
+import ChartTemplate from '../ChartTemplate';
+
+const meta: Meta = {
+    title: 'Chart/RadialCharts',
+};
+const df = [
+    { x: 'Alpha', y: 100 },
+    { x: 'Beta', y: 50 },
+    { x: 'Gamma', y: 20 },
+    { x: 'Delta', y: 30 },
+];
+export default meta;
+
+export const PieChart = ChartTemplate.bind({});
+const basePieChartArgs: Props<DataFrame, ChartOptions> = {
+    data: {
+        loading: false,
+        value: df,
+    },
+    options: {
+        labelColumn: 'x',
+        title: {
+            text: 'Pie Chart',
+        },
+        series: [
+            {
+                type: ChartSeriesType.Pie,
+                valueColumn: 'y',
+                backgroundColor : ['#CB4335', '#1F618D', '#F1C40F', '#27AE60']
+            },
+        ],
+    },
+};
+PieChart.args = basePieChartArgs;
+
+export const PieDataLabelCustomLegendValues = ChartTemplate.bind({});
+const PieDataLabelCustomLegendValuesArgs: Props<DataFrame, ChartOptions> = {
+    data: {
+        loading: false,
+        value: df,
+    },
+    options: {
+        labelColumn: 'x',
+        source: defaultSource,
+        series: [
+            {
+                type: ChartSeriesType.Pie,
+                valueColumn: 'y',
+                backgroundColor: ['#CB4335', '#1F618D', '#F1C40F', '#27AE60'],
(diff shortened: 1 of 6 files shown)


packages/visualizations-react/stories/Chart/RadialCharts/RadialCharts.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+        ],
+    },
+    options: {
+        labelColumn: 'x',
+        title: {
+            text: 'Pie Chart',
+        },
+        series: [
+            {
+                type: ChartSeriesType.Pie,
```
KevinFabre-ods, line 61: IMO, we need to set the type to `ChartSeriesType.Doughnut` so that even if the cutout option is not defined the chart will have a default cutout of '50%'

https://www.chartjs.org/docs/latest/charts/doughnut.html#doughnut-and-pie-charts

packages/visualizations/src/components/Chart/types.ts
```diff
@@ -39,6 +39,8 @@ export interface ChartOptions {
     description?: string;
     /** Link button to source */
     source?: Source;
+    /** Cutout percentage for doughnut charts */
+    cutout?: string;
```
KevinFabre-ods, line 43: Maybe we can move this option in a sub object. IMO, it should not be on the top level of options as this is specific to pie and doughnut charts. 

We can handle this later when implementing other new charts.

KevinFabre-ods (changes requested):
GG, you did well 👏 

I left minors comments 😉

etienneburdet (changes requested):
LGTM. A bit sad for the TS bug, but I couldn't find anything better either (it's not that bad once you know there is bug).

We need to set Lerna in "independant mode though", in the mean time, I would just reset the `api-client` files, so that they aren't patched.

KevinFabre-ods:
> We need to set Lerna in "independant mode though"
https://github.com/opendatasoft/ods-dataviz-sdk/blob/b4822a16fbf930c2c512afa45f99bbd5481d10e3/lerna.json#L5

It's already the case. 

>  I would just reset the api-client files, so that they aren't patched.

I tried to remove the api client changes, but every time I do `lerna version --no-push` or `lerna version --no-push --conventional-graduate` lerna wants to update the version of `@opendatasoft/api-client`. 

I think this commit https://github.com/opendatasoft/ods-dataviz-sdk/commit/b4822a16fbf930c2c512afa45f99bbd5481d10e3 used in a [squash merge](https://github.com/opendatasoft/ods-dataviz-sdk/pull/166) broke some of lerna's mechanisms for detecting changes in individual packages.

It may be that we don't have much choice left and we'll have to accept a `@opendatasoft/api-client` 21.0.1 version to solve this problem.

packages/visualizations-react/stories/Chart/StudioLayouts/DoughnutChart.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
+            },
+        },
+        title: {
+            text: 'Doughnut chart with title and legend with values',
+        },
+    },
+};
+DoughnutTitleLegendValues.args = DoughnutTitleLegendValuesArgs;
+
+export const DoughnutDataLabelCustomLegendValues = ChartTemplate.bind({});
```
KevinFabre-ods, line 148: This is not a Studio layout, this should be moved in the RadialCharts stories

packages/visualizations-react/stories/Chart/RadialCharts/RadialCharts.stories.tsx
AnthoPepin: You're right, and then the `cutout` option belongs to the serie which makes more sense => if we have a `Doughnut` serie then we have a cutout in this serie, else we don't

KevinFabre-ods (approved):
LGTM 👏

KevinFabre-ods:
@etienneburdet We reached our monthly limit on Chromatic. Before merging this PR, could you please do a quick tour of the Storybook ? 

Thanks 🙏

PR #181, opened 2023-08-08
POC: factor map component

etienneburdet:
Simple refactor POC, only ?path=/story/map-choroplethvector--studio-choropleth-vector-gradient actually works. Will stay in draf until we are more clear on POI.

The main goal is to split `MapRender` into smaller decoupled "parts" (components, functions etc.), so that it's easier to add/remove stuff.

`MapCard` handles whatever  isn't Maplibre: title, description, layout of buttons etc.

`Map` is a basic map that takes no props specific to a kind of map (Choropleth, geojson) etc. and instanciate a map that resizes. It exposes a new `load` event to access all the methods we want to (setBbox, addControl, addLayer, addSource etc.), so that `Choropleth` or `Poi` can do whatever they want based on their props.

## Other options
The other method to expose `map.xxx` or methods wrapping it could be to use context: 
```
<Map>
  <Sources>
  {each layers as layer}
    <Layer layer={layer} />
  {/each}
  <Styles>
  <Control>
</Map>
```
Where map uses `setContext(key, { publicInterface}` and  child components use `getContext` to retrieve map and/or methods (very much like the [Svelte tutorial](https://svelte.dev/tutorial/context-api)).

This second options implies components with only JS. They are a bit strange because their order doesn't matter, but they have access to lifeCycle functions.

Why not? :shrug:

## Discussion
If we move to `setStyles` (which seems great), do we expose `setStyle`  or should `Map` get a `style` props ? This seems reasonable and have more of a "downward" flow.


KevinFabre-ods:
Thanks

Took some part of the Map.ts code to isolate MapLibre dependency for  POI Map component https://github.com/opendatasoft/ods-dataviz-sdk/pull/185/commits/71cf15c88fa9ac905eacfa550379d4ebd9a35cb1

We will do a large refactoring (POI Map + choropleth) when we have something dry

KevinFabre-ods:
Thanks

Took some part of the Map.ts code to isolate MapLibre dependency for POI Map component https://github.com/opendatasoft/ods-dataviz-sdk/commit/71cf15c88fa9ac905eacfa550379d4ebd9a35cb1

We will do a large refactoring (POI Map + choropleth) when we have something dry

PR #182, opened 2023-08-11
[FEATURE][SC-42061] Add scatter chart

AnthoPepin:
## Summary

The goal for this PR is to add a new type of chart: [Scatter](https://www.chartjs.org/docs/latest/charts/scatter.html)

⚠️  **This is a very simple implementation, which forms the basis for iterating according to future needs.**

⚠️  Commits will be purged before merge

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-42061](https://app.shortcut.com/opendatasoft/story/42061).

Screenshot 2023-09-08 at 11 37 27

### Changes

- Update `NumericCartesianAxisConfiguration` types to allow linear Cartesian axis to begin at zero
   **x and y axis doesn't begins at zero by default (so no breaking changes)**
- Handle the scatter case for tooltip and dataset builds

#### Breaking Changes

none

### Changelog

Add new chart type: scatter

## Review checklist

- [x] Description is complete

### DIFF
--- a/packages/visualizations/src/components/Chart/scales.ts
+++ b/packages/visualizations/src/components/Chart/scales.ts
@@ -81,6 +81,9 @@ export default function buildScales(options: ChartOptions): ChartJsChartOptions[
     // X Axis
     if (options.axis?.x) {
         scales.x = {
+            ...(options.axis.x.type === 'linear' && {
+                beginAtZero: defaultValue(options?.axis?.x?.beginAtZero, true),
+            }),
             stacked: options.axis?.assemblage?.stacked,
             max:
                 options?.axis?.x?.type === 'linear' && options.axis?.assemblage?.percentaged
@@ -130,6 +133,9 @@ export default function buildScales(options: ChartOptions): ChartJsChartOptions[
     // Y Axis
     if (options.axis?.y) {
         scales.y = {
+            ...(options.axis.y.type === 'linear' && {
+                beginAtZero: defaultValue(options?.axis?.y?.beginAtZero, true),
+            }),
             stacked: options.axis?.assemblage?.stacked,
             max:
                 options?.axis?.y?.type === 'linear' && options.axis?.assemblage?.percentaged
--- a/packages/visualizations-react/stories/Chart/AxisAssemblages/AxisAssemblages.stories.tsx
+++ b/packages/visualizations-react/stories/Chart/AxisAssemblages/AxisAssemblages.stories.tsx
@@ -776,7 +776,7 @@ function randomNormal(mean = 0, stdDev = 1) {
     return z0 * stdDev + mean;
   }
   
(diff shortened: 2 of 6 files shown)


packages/visualizations/src/components/Chart/scales.ts
```diff
@@ -81,6 +81,9 @@ export default function buildScales(options: ChartOptions): ChartJsChartOptions[
     // X Axis
     if (options.axis?.x) {
         scales.x = {
+            ...(options.axis.x.type === 'linear'
+                ? { beginAtZero: defaultValue(options?.axis?.x?.beginAtZero, false) }
+                : {}),
```
RafaelSzmarowski, line 86:
```suggestion
            ...(options.axis.x.type === 'linear' && { beginAtZero: defaultValue(options?.axis?.x?.beginAtZero, false) }),
```

Maybe a more compact way to write it, should have the same output ?

RafaelSzmarowski (approved):
Nice work, well done ! LGTM, I left a really minor comment, just a matter of style so not mandatory at all.

wmai (approved):
Nicely done! LGTM 👏

KevinFabre-ods:
I think I added a regression with https://github.com/opendatasoft/ods-dataviz-sdk/pull/182/commits/e431421be6265ddb5f30017f31d3fab7923bd639

When reviewing the change on Chromatic, we can see that some axis do not longer begin at zero.

For scatter charts, axis begin at the lowest value (so not zero).

Screenshot 2023-09-18 at 16 46 15


If some values are negative, axis will begin at the lowest negative value, even if `beginAtZero` is `true`.


image


What do you think @RafaelSzmarowski & @wmai ? We should  still expose `beginAtZero` and only add it in the `buildScale` function if defined.

```
...(options.axis.x.type === 'linear'  && isBoolean(options?.axis?.x?.beginAtZero) && {beginAtZero: options?.axis?.x?.beginAtZero}), 
```

https://github.com/opendatasoft/ods-dataviz-sdk/blob/c4c637d3bbbb76ca431edefb750d5d524c31795a/packages/visualizations/src/components/Chart/scales.ts#L85

KevinFabre-ods:
@AnthoPepin  
In https://github.com/opendatasoft/ods-dataviz-sdk/pull/182/commits/7ca0df5208e8ce11c5b50b4b6760b9a1cf05f2da, One Scatterplot story has its snapshot been disabled for Chromatic. As it uses random values, the visual representation of the chart changes at every new render, so there is no gain to save a snapshot of this story 😉

AnthoPepin:
Sorry about that, the goal of those stories was studying the case of high density and try out some configuration.

KevinFabre-ods:
> Sorry about that, the goal of those stories was studying the case of high density and try out some configuration.

> No worries, it's just to keep you up to date on how your work has been taken over.

PR #183, opened 2023-08-14
Add /catalog/datasets/{dataset_id}/records/{record_id} endpoint and a test

fpassaniti:
## Summary

The goal for this PR is to add the missing endpoint for datasets/x/records/y/

### Changes

Just the addition of a line to build the corresponding query

### Changelog

In the Dataset object: It's now possible to build a query to /catalog/datasets/{dataset_id}/records/{record_id}"
[See the documentation of this endpoint for more information](https://help.opendatasoft.com/apis/ods-explore-v2/#tag/Dataset/operation/getRecord)

### Example 

`fromCatalog().dataset('my_dataset').record('a_record_id').toString())` will return `catalog/datasets/my_dataset/records/a_record_id/`

PR #184, opened 2023-08-30
feat: Add Record endpoint in api-client

richterb:
## Summary

The goal for this PR is to catch-up the merged "support record endpoint in api-client" PR (#183) with conventional commits.

### Changes

I added a small bit of doc, which fits what currently exists in the README. It's starting to look a bit "light" if we consider this a doc, but we'll see that later.
I edited the PR template to mention conventional commits.

## Open discussion
It would be nice indeed to only offer ODSQL features (`.where()`, `.limit()` etc) on endpoints that support it, but this doesn't seem trivial at all, especially if we want to do it right and really support the right parameters on each endpoint. I think it's firmly out of scope for this one.

## Review checklist

- [ ] Description is complete

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -35,6 +35,7 @@ _special cases that are not widely known_
 ## Review checklist
 
 - [ ] Description is complete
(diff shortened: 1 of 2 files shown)


KevinFabre-ods (approved):
LGTM 

🙏

PR #185, opened 2023-09-01
[FEATURE][SC-42583] Hand all kind of sources for POI Map

KevinFabre-ods:
## Summary

The goal for this PR is to update props that are passed to the POI Map component

This is the result of a [POC](https://github.com/opendatasoft/ods-dataviz-sdk/pull/178)

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-42583](https://app.shortcut.com/opendatasoft/story/42583).

### Changes

Our ODS usage of the POI maps requires us to be able to changes sources and layers very often.
Sources and layers can be set in the [MapLibre style object](https://maplibre.org/maplibre-style-spec/).
So in this implementation, whenever a source or a layer changes, the full style object is rebuilt and the map updated with map.setStyle(style).

#### Technical changes

A Map class has been created to move all function related to MapLibre in a isolated file. Making the MapRender component cleaner.

#### Breaking Changes

none

## Open discussion

I didn't find a simple way to center the map bounding box to the rendered features.

This can be done with a lot of code by using map event and getRenderedFeatures. This would be much butter if we could rely on the `bbox` prop 

## Review checklist

- [x] Description is complete


RafaelSzmarowski (approved):
LGTM ! Nice work and nice clean Map renderer. To add my point of view on the open discussion. For now we can keep `bbox` as an option passed to the component but it would be a nice add-on in the future to handle it directly from the maplibre source if possible.

packages/visualizations/src/components/MapPoi/utils.ts
```diff
@@ -1,31 +1,60 @@
...
+            paint: {
+                'circle-radius': 5,
+                // TODO: Better typing
+                'circle-color': cirleColor as DataDrivenPropertyValueSpecification<string>,
+            },
+            filter: ['==', ['geometry-type'], 'Point'],
+        };
     });
-    matchExpression.push(noMatchColor || DEFAULT_COLORS.Blue); // Default fallback color
+};
```
wmai, line 51: What do you think of this typing for the `circleColor` variable? (There's a small typo btw, ~cirleColor~ 👉 circleColor)

``` TS
// ...

import type { Color } from '../types';

// ...

// Only circle layers are supported
export const getMapLayers = (layers?: Layer[]): CircleLayerSpecification[] => {
    if (!layers) return [];

    return layers.map((layer) => {
        let circleColor: DataDrivenPropertyValueSpecification<Color> | Color = layer.color;
        if (layer.colorMatch) {
            const { key, colors } = layer.colorMatch;
            const groupByColors = ['match', ['get', key]];
            Object.keys(colors).forEach((color) => {
                groupByColors.push(color, colors[color]);
            });
            groupByColors.push(layer.color);
            circleColor = groupByColors;
        }
        const { id, type, source, sourceLayer } = layer;
        return {
            id,
            type,
            source,
            ...(sourceLayer ? { 'source-layer': sourceLayer } : undefined),
            paint: {
                'circle-radius': 5,
                'circle-color': circleColor,
            },
            filter: ['==', ['geometry-type'], 'Point'],
        };
    });
};
```

wmai (approved):
LGTM! Only one very small suggestion regarding the typing.

packages/visualizations/src/components/MapPoi/utils.ts
KevinFabre-ods: Thanks, that's much better 💪

PR #186, opened 2023-09-05
[FEAT][PoiMap][SC-43161] Show popup on feature click

KevinFabre-ods:
## Summary

The goal for this PR is to show popup on the map after a feature has been clicked.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-43161](https://app.shortcut.com/opendatasoft/story/43161).

Screenshot 2023-09-05 at 18 01 21

### Changes

Poi map can now display popup when a feature is clicked.

To do so, a popup config object must be defined in a layer.
```js
{
 ...layerConfig, 
 popup: {
    display: 'sidebar', // or 'tooltip'
    getContent: (id: GeoJSONFeature['id'], properties?: GeoJsonProperties) => 'my popup content'; 
   }
}
```
The SDK provides only a skeleton (positioning and content container style) for the popup.
Popup content comes from the result of `getContent`.

Also when interaction is switch off: 
- The view box is reset to the bbox
- The popup is closed

#### Breaking Changes

none

## To be tested

- Try to use multiples sources (from different types) and layers
- Try to used asynchronous calls to define the popup content 
- All props changes should be taken into consideration by the map

## Review checklist

- [x] Description is complete


KevinFabre-ods:
@RafaelSzmarowski and @wmai Sorry, I didn't realize my latest rebase broke the commit history. 

Now this PR also only changes related to the popup 😉

packages/visualizations/src/components/MapPoi/Map.ts
```diff
@@ -2,46 +2,155 @@ import type { BBox } from 'geojson';
...
+        }
+
+        this.popup.setHTML(getLoadingContent());
+
+        getContent(featureId, properties).then((content) => {
+            this.popup.setHTML(content);
+        });
+
+        const classnameModifier = display === 'sidebar' ? 'addClassName' : 'removeClassName';
+        this.popup[classnameModifier](`${POPUP_OPTIONS.className}--as-sidebar`);
```
etienneburdet, line 134: Wouldn't it be easier to create our own display in this case? And decide in svelte wether we want to display our own or generate a maplibre popup. This would make a useful component I think (and scope css a bit more).

But this works well and is quite clean, so if you feel like it's not worth the hassle, I can understand.

etienneburdet (commented):
Looks really clean! The map and maprender implem look really nicer, easier to read and more stable.

There are just two little things I'd rather do without: the queue mechanism and the `.bind(this)`. This are pattern I'm not sure everybody in the future will know. Moving to a Svelte component should solve that, while leveraging 99% of this code (this is for another PR).

The popup I feel like could be done as a mix of maplibre'spopup or HTML (especially if we want to support more "sidebar" stuff, like next to the map).

Anyways, these comments are a bit more far fetch and doesn't block merge imo. I'll approve if you want to leave it like this.

packages/visualizations/src/components/MapPoi/Map.ts
KevinFabre-ods: Yes, I think we can all agree that a `MapLibre.Popup` to manage the behaviour of the sidebar is not the best option. 

At the moment, there's no clear direction in which the design wants to go with this sidebar.

IMO, we could keep it as it is. It's a simple implementation and can easily be removed later.

KevinFabre-ods:
> There are just two little things I'd rather do without: the queue mechanism and the .bind(this). This are pattern I'm not sure everybody in the future will know. Moving to a Svelte component should solve that, while leveraging 99% of this code (this is for another PR).

👍 
Let's make a ticket where we save all technical / refactor ideas we have for the Map components

wmai (approved):

packages/visualizations/src/components/MapPoi/Map.ts
```diff
@@ -59,6 +168,10 @@ export default class MapPOI {
      * - `baseStyle` could be removed
      * - The key block could also be removed in MapRender.svelte
      */
+    /**
+     * Update the sources and layers of the map.
+     * Layers of the map base style are untouched.
+     */
```
RafaelSzmarowski, line 174: I guess this function should clean the opened popups and popupFeatures array as they are associated to a specific source that can change here. I have had a bug while editing a map when I have clicked to open a tooltip, leave it opened and then changed the source.

Adding : 
```
this.popup.remove();
this.popupFeatures = [];
```

Do the trick for my case but probably we will want something cleaner at some point and more general, what are the other things to clean when changing the source ? layerIds ?

KevinFabre-ods: Should be good now with https://github.com/opendatasoft/ods-dataviz-sdk/pull/186/commits/8a46de55acbbba197f4482669c3e8f8ae6c6bf15

Please test it with a local version of the SDK

packages/visualizations/src/components/MapPoi/Map.ts
```diff
@@ -2,46 +2,155 @@ import type { BBox } from 'geojson';
...
+
+        if (this.popup.isOpen() === false) {
+            this.popup.setLngLat(geometry.coordinates.slice() as LngLatLike).addTo(map);
+        }
+
+        this.popup.setHTML(getLoadingContent());
+
+        getContent(featureId, properties).then((content) => {
+            this.popup.setHTML(content);
+        });
```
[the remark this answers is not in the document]
KevinFabre-ods: IMO, we should do some fine tuning in the Studio. 

If no fields has been chosen in the Studio for title, image and content, we could display a placehoder tooltip to show that something will be display once title, image or content fields are chosen.

In the SDK we already have a condition to show popup: the layer must have a `popup` key. 


What if `getLoadingContent` is defined and `getContent` returns an empty string ? 
the popup show loading... and then disappear ? 
Maybe we should handle the Catch case to display an onErrorContent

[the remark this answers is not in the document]
etienneburdet: I'm OK with popup being rendered as long as there is a `popup` key. IMO it's the Studio/PageRenderer responsability to no send an empty popup, but there are case we might want to support empty strings in the SDK (e.g. the render function depends on a fetch and is empty before that).

packages/visualizations/src/components/MapPoi/Map.ts
RafaelSzmarowski: Thanks ! It seems to work as expected now on source changes, but I still have the same error when removing the map instance (for example, you have a configured map, open a tooltip, and then change the block type while the tooltip is open) It seems that the popup is not removed in the map.remove function ! 

Adding line 170 this code fixes it: 
```
remove() {
        this.popup.remove();
        this.popupFeatures = [];
        this.queue((map) => map.remove());
    } 
```

packages/visualizations/src/components/MapPoi/Map.ts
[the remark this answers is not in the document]
RafaelSzmarowski: Well indeed it seems like it should be the SDK part to handle empty tooltips content ! Let's keep it like this !

packages/visualizations/src/components/MapPoi/Map.ts
KevinFabre-ods: ah yes, I didn't test when the map is destroyed, thanks !

RafaelSzmarowski (approved):
LGTM ! Nice work

KevinFabre-ods:
Before merge I'll do a squash to keep only two commits: 
- a feat commit [feat(PoiMap): can display a popup on feature click](https://github.com/opendatasoft/ods-dataviz-sdk/pull/186/commits/212d04464f980d9aff1dfcb216ed49b787d6d96b)
- and the chore: publish new versions (that will be drop when https://github.com/opendatasoft/ods-dataviz-sdk/pull/164) is merged in `main`

PR #188, opened 2023-09-11
Implement treemap

JulietteRevOpendatasoft:
## Summary

This PR has been done in peer coding with @RafaelSzmarowski :pray:
The goal for this PR is to implement the treemap chart in the sdk (and then to use it on the new Data page in the platform: [figma](https://www.figma.com/file/qEguDqjQElqx542a3Ivevv/Domain-business-dashboard?type=design&node-id=3219-12804&mode=design&t=zY37c8r9qGDfc4j9-0)).

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-43363](https://app.shortcut.com/opendatasoft/story/43363/update-data-page-use-treemap-for-theme-distribution).

### Changes
use of `chartjs-chart-treemap` library.
use Treemap as a new Chart type.

## Open discussion
I couldn't find a way to create something as the `MiddleEllipsis` in the platform to avoid that:
image
I think it should be on SDK side, but I might be wrong ? You may already have something like this in SDK ?

I tried to implement a tooltip that displays only the category count. I am not sure it can be done, please let me know if you find a way to do it.

## Review checklist

- [ ] Description is complete


packages/visualizations/src/components/utils/formatter.ts
```diff
@@ -22,7 +22,7 @@ export function defaultCompactLegendNumberFormat(value: number): string {
 }
 
 export function assureMaxLength(value: string, maxLength: number) {
-    if (value.length > maxLength) {
+    if (value && value.length > maxLength) {
```
etienneburdet, line 25:
```suggestion
    if (value?.length > maxLength) {
```

etienneburdet: I'm surprised we don't have a rule for that 🤔

packages/visualizations/package-lock.json
etienneburdet, line 1: Looks like was compiled under node 14, I'll rebuild that.

etienneburdet (approved):
Looking good to me! I was afraid the optional `labelColumn` could create uncatched cases on the platform side, but it doesn't seem so.

I've just prettier'ed some files and rebuild the `.lock` with node 16.

packages/visualizations/package-lock.json
RafaelSzmarowski: Normally we finally pushed the node 16 version! The PR is not yet totally finished as we still have some typescript `any` left. Don't know if you want to take a look on that, if not I'll try to do it later 😃

packages/visualizations/src/components/utils/formatter.ts
RafaelSzmarowski: You're right about the suggestion in this case but I have the feeling that such rule will be hard to enforce in all situations, in some cases optional chaining might not be recommended as optional chaining and AND operator don't give always the same output ?

RafaelSzmarowski:
> I couldn't find a way to create something as the `MiddleEllipsis` in the platform to avoid that:  I think it should be on SDK side, but I might be wrong ? You may already have something like this in SDK ?
> 
> I tried to implement a tooltip that displays only the category count. I am not sure it can be done, please let me know if you find a way to do it.

Indeed, this display is not good. The treemap plugin only allows choice between '`cut`', '`fit`', '`hidden`'. It's hard to implement a proper and smart `MiddleEllipsis` inside a `canvas`, maybe I'm missing something but it would have require a custom plugin to detect the size of the box and if the text touches the limits and probably many other variables... The solution, that we also use in other charts, is to decide of a `maxLength` for the labels (eg: 10), keep the fit property as the labels variation shouldn't be huge, and to display the whole label inside the tooltip so the whole information is still accessible: 
Capture d’écran 2023-09-28 à 14 22 00

As for the tooltip, `chartjs` allows to do pretty much anything, in this case we had to change specifically for the `Treemap` some additional default tooltip callbacks `title` and remove the color box through `displayColors` props.

RafaelSzmarowski (approved):
The `Treemap` is a very specific chart that differs from what we usually did. We've tried to keep the implementation close to our choices on the SDK from the beginning, force some behaviour to make less options available for users but a nice final display. But this way, we had to add a few `if` conditions (like for example in Chart.svelte in the tooltip callbacks). Also as the `Treemap` plugin doesn't seem to really support typescript, we also had to add some custom types and cast them in some places as it was hard for typescript to properly infers them. The resulting code might look a little bit messy but it sticks to what was made previously (Doughnut and ScatterPlot)

I have the general feeling that we might need some time to rethink some parts of the implementation and how we want to handle next chart types. We shouldn't do it in this PR to not hold longer this chart type but probably before the next one as this one offers us a good vision of how a different chart fits inside our code structure. What do you think @wmai  ?

RafaelSzmarowski:
Also, I don't know why my `package.lock` indentation differs from the `main` file. Any idea @wmai ?

wmai (commented):
Work like a charm, very well done! 💪 Thank you for your help on this PR.

> The Treemap is a very specific chart that differs from what we usually did. We've tried to keep the implementation close to our choices on the SDK from the beginning, force some behaviour to make less options available for users but a nice final display. But this way, we had to add a few if conditions (like for example in Chart.svelte in the tooltip callbacks).

Regarding the design choices you made in the Treemap, everything looks good to me. Given the constraints and architecture of the current SDK, your approach seems well-aligned.

> Also as the Treemap plugin doesn't seem to really support typescript, we also had to add some custom types and cast them in some places as it was hard for typescript to properly infers them. The resulting code might look a little bit messy but it sticks to what was made previously (Doughnut and ScatterPlot)

About the TypeScript support for Treemap, I noticed that the plugin does have TypeScript definitions available [here](https://github.com/kurkle/chartjs-chart-treemap/blob/main/types/index.esm.d.ts). Can you elaborate on why you think it doesn't support TypeScript?


> Also, I don't know why my package.lock indentation differs from the main file. Any idea @wmai ?

For the issue with `package.lock` indentation, I've found that running `npm run build-package-lock` using Node version `16.20.2` resolves the issue on my local environment. Could you try that and see if it works for you?

RafaelSzmarowski:
> For the issue with `package.lock` indentation, I've found that running `npm run build-package-lock` using Node version `16.20.2` resolves the issue on my local environment. Could you try that and see if it works for you?

Unfortunately, it doesn't seem to work, I've tried to rebuild the package-lock using different Node versions (16.X, 17) but the indentation still remains changed as in this branch. It must be coming from a config file but I can't find it ! Even replacing the file from `main`, committing and rebuilding changes the indentation !

wmai:
> > For the issue with `package.lock` indentation, I've found that running `npm run build-package-lock` using Node version `16.20.2` resolves the issue on my local environment. Could you try that and see if it works for you?
> 
> Unfortunately, it doesn't seem to work, I've tried to rebuild the package-lock using different Node versions (16.X, 17) but the indentation still remains changed as in this branch. It must be coming from a config file but I can't find it ! Even replacing the file from `main`, committing and rebuilding changes the indentation !

When I try this command, `apiclient` and `visualizations-react` indentations are updated (which is good). Do you have the same results?

RafaelSzmarowski:
> > > For the issue with `package.lock` indentation, I've found that running `npm run build-package-lock` using Node version `16.20.2` resolves the issue on my local environment. Could you try that and see if it works for you?
> > 
> > 
> > Unfortunately, it doesn't seem to work, I've tried to rebuild the package-lock using different Node versions (16.X, 17) but the indentation still remains changed as in this branch. It must be coming from a config file but I can't find it ! Even replacing the file from `main`, committing and rebuilding changes the indentation !
> 
> When I try this command, `apiclient` and `visualizations-react` indentations are updated (which is good). Do you have the same results?

I haven't done this for `apiclient` and `visualizations-react` as they shouldn't change after my changes ? Only for `visualizations` which changes the indentation for a bigger one on this branch than the one in the `main` branch

PR #194, opened 2023-09-27
[FEATURE][POI MAP] Add `zoom`and `center` options

KevinFabre-ods:
## Summary

The goal for this PR is to add `zoom`and `center`in the options prop

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-43528](https://app.shortcut.com/opendatasoft/story/43528).

### Changes

Fix a behavior where the map bbox was updated when a change in `options`. Did the same thing for the center prop.In `Map.svelte` the `bbox` and `center` given to `mapRender` change only their value change not their reference. 

Maybe we can factorize this code into something like` const freshCenter = useFreshValue(center, comparaisonCallback)`

#### Breaking Changes

none

### Changelog

/

## Open discussion

/

## To be tested

/

## Review checklist

- [x] Description is complete

### DIFF
--- a/packages/visualizations/src/components/MapPoi/Map.svelte
+++ b/packages/visualizations/src/components/MapPoi/Map.svelte
@@ -35,10 +35,10 @@
         interactive,
     } = getMapOptions(options));
 
-    let bbox = createDeepEqual(_bbox);
-    let center = createDeepEqual(_center);
-    $: bbox.set(_bbox);
-    $: center.set(_center);
+    const bbox = createDeepEqual(_bbox);
+    const center = createDeepEqual(_center);
+    $: bbox.update(_bbox);
+    $: center.update(_center);
 </script>
 
 <div>
--- a/packages/visualizations/src/components/MapPoi/Map.svelte
+++ b/packages/visualizations/src/components/MapPoi/Map.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-    import type { BBox } from 'geojson';
+    import createDeepEqual from '../../stores/createDeepEqual';
 
     import MapRender from './MapRender.svelte';
 
@@ -12,16 +12,10 @@
         getPopupsConfiguration,
         getMapOptions,
     } from './utils';
-    import type { Center, PoiMapData, PoiMapOptions } from './types';
+    import type { PoiMapData, PoiMapOptions } from './types';
 
     export let data: Async<PoiMapData>;
     export let options: PoiMapOptions;
-
-    let bbox: BBox;
-    let previousBbox: BBox;
-
-    let center: Center;
-    let previousCenter: Center;
 
     $: style = getMapStyle(options.style);
     $: sources = getMapSources(data.value?.sources);
@@ -29,9 +23,9 @@
     $: popupsConfiguration = getPopupsConfiguration(data.value?.layers);
 
     $: ({
-        bbox: currentBbox,
+        bbox: _bbox,
         zoom,
-        center: currentCenter,
+        center: _center,
         title,
         subtitle,
         description,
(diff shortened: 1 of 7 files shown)


packages/visualizations/src/components/MapPoi/types.ts
```diff
@@ -21,6 +24,8 @@ export interface PoiMapOptions {
      * Also set the position of the map when rendering.
      */
     bbox?: BBox;
+    center?: Center;
```
etienneburdet, line 27: `LngLatLike` from `maplibre` I think

etienneburdet (changes requested):
The whole concept seems pretty good. I'm really in favor of factoring the zoom/bbox reactivity into its own thing.

As for exposing zoom/center separately or as one prop, it's a bit stylistic… I kinda prefer the first, but the second is flatter. Honestly, doesn't make that much differenc.

packages/visualizations/src/components/MapPoi/Map.svelte
```diff
@@ -10,22 +12,78 @@
...
+    $: {
+        if (
+            currentCenter &&
+            (!previousCenter ||
+                currentCenter.some((point, index) => point !== previousCenter[index]))
+        ) {
+            center = currentCenter;
+            previousCenter = currentCenter;
+        }
+    }
```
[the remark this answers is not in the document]
KevinFabre-ods: What do you think of this https://github.com/opendatasoft/ods-dataviz-sdk/pull/194/commits/b3cf1f3f5ac212ffa9dc0cc52a912f8b376f3e82 ?

KevinFabre-ods:
> The whole concept seems pretty good. I'm really in favor of factoring the zoom/bbox reactivity into its own thing.
> 
> As for exposing zoom/center separately or as one prop, it's a bit stylistic… I kinda prefer the first, but the second is flatter. Honestly, doesn't make that much differenc.

I used `jumpTo` in https://github.com/opendatasoft/ods-dataviz-sdk/pull/194/commits/8ef37e99d79a4b5143a69473fe795226d121f473

packages/visualizations/src/stores/createDeepEqual.ts
```diff
@@ -0,0 +1,29 @@
...
+         */
+        set: (newValue: V | undefined) => {
+            if (JSON.stringify(newValue) !== JSON.stringify(get(value))) {
+                value.set(newValue);
+            }
+        },
+    };
+};
+
+export default createDeepEqual;
```
etienneburdet, line 29:
```suggestion
import { writable, get } from 'svelte/store';
import { isEqual } from 'lodash';
/**
 * Creates a Svelte writable store that compares values deeply before updating.
 *
 * @param initialValue - The initial value for the store.
 * @returns An object containing the subscribe and set methods.
 */
const createDeepEqual = <V>(initialValue: V | undefined) => {
    const { subscribe, set} = writable<V | undefined>(initialValue);

    return {
        /**
         * Subscribes to changes in the store's value.
         */
        subscribe,
        set,
        /**
         * Sets the new value for the store if it differs from the current value
         * by performing a deep comparison.
         */
        update: (value: V, newValue?: V) => {
            if (isEqual(value, newValue)) {
                return newValue;
            }
            return value;
        },
    };
};

export default createDeepEqual;
```
I think using update is a bit more efficient, since it avoids reading the whole store in `set`. Also, in the long term `isEqual` is both more efficient and also covers objects, Sets and Maps with different ordering etc.

etienneburdet (changes requested):
Looks super clean, I really like the idea of the deepEqual store in the long run. 

Just added a few comments on the store itself, but that does basically the same thing.

packages/visualizations/src/stores/createDeepEqual.ts
KevinFabre-ods: Thx, I used store.update() to get the current store value https://github.com/opendatasoft/ods-dataviz-sdk/pull/194/commits/f8f0f37163ab9dfb0c82bee2f60f77ccf95d327b

packages/visualizations/src/stores/createDeepEqual.ts
```diff
@@ -7,21 +9,23 @@ import { writable, get } from 'svelte/store';
...
         /**
-         * Sets the new value for the store if it differs from the current value
+         * Update the store value if the new value differs from the store current value
          * by performing a deep comparison.
          */
-        set: (newValue: V | undefined) => {
-            if (JSON.stringify(newValue) !== JSON.stringify(get(value))) {
-                value.set(newValue);
-            }
+        update: (newValue?: V | undefined) => {
```
etienneburdet, line 23: Ha, very clever use of update!

etienneburdet (approved):
All good for me, looking nice!

PR #196, opened 2023-09-29
[FIX][POI Map] Clicked feature is always visible when sidebar popup is open

KevinFabre-ods:
## Summary

The goal for this PR is to add a simple rule to make the clicked feature still visible when the sidebar popup is open. 

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-43611](https://app.shortcut.com/opendatasoft/story/43611).

### Changes

If the popup is of type 'sidebar', the map will be centred on the element clicked, with a left margin of 300px.
As you can see in the video, we don't return to the initial center before clicking. 

IMO, this is still a good improvement and as the sidebar is probably going to be completely redesigned, it may not be worth spending any more time on it to try to going back to the original center.

https://github.com/opendatasoft/ods-dataviz-sdk/assets/117300300/22045473-425b-453c-96a9-e400721eab3f

#### Breaking Changes
/

### Changelog

/

## Open discussion

/

## To be tested

/

## Review checklist

- [x] Description is complete

### DIFF
--- a/packages/visualizations/src/components/MapPoi/Map.ts
+++ b/packages/visualizations/src/components/MapPoi/Map.ts
@@ -10,7 +10,7 @@ import maplibregl, {
     StyleSpecification,
 } from 'maplibre-gl';
 
-import { POPUP_OPTIONS } from './constants';
+import { POPUP_OPTIONS, POPUP_WIDTH } from './constants';
 import type { PopupsConfiguration, CenterZoomOptions } from './types';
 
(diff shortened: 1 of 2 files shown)


etienneburdet (approved):
All good for me. It's temporary fix indeed—we will certainly want to something around zooming at some point too—but a good improvement that remains cheap for us.

I'll give a think about what kind of API we want to expose for the zooming/centering part.

PR #203, opened 2023-11-13
[SC-44672][FEAT][Poi Maps] Support 'symbol' layers

KevinFabre-ods:
## Summary

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-44672](https://app.shortcut.com/opendatasoft/story/44672).

The goal of this PR is to support a new type of layer for POI maps.

The need comes from our product roadmap where we offer our customers the ability to customise a marker.

The marker customization concerns the color, the icon and the color of the icon.

An icon could be something like this:
image

We made the decision to rely on a `symbol` to display this type of marker.

### Changes
- add a new reactive option: `images` The map will automatically loads images (via URL). Images no longer defined are also automatically pruned.
- support a new layer type: `symbol`. Layer generation function have been updated to handle the two types ('circle' and 'symbol')

#### Breaking Changes

none

## Open discussion

none

## To be tested

Test it with new story

## Review checklist

- [x] Description is complete


RafaelSzmarowski:
In Storybook I have a console warning: 
Capture d’écran 2023-11-23 à 14 34 58
From my understanding, it's because the image load is async and the the layer is drawn before the image is loaded. Then the image is loaded and appears on the map. In the documentation it seems that they first load and add images then they add sources and layers (https://maplibre.org/maplibre-gl-js/docs/examples/add-image/). 

I don't know if there is a much bigger risk to do the images loading and layer drawing separately and that it's only a minor and temporary console warning but if we want to follow the documentation example, the image load should probably be inside `setSourcesAndLayers`. What do you think, maybe you already tested that ?

KevinFabre-ods:
> In Storybook I have a console warning: Capture d’écran 2023-11-23 à 14 34 58 From my understanding, it's because the image load is async and the the layer is drawn before the image is loaded. Then the image is loaded and appears on the map. In the documentation it seems that they first load and add images then they add sources and layers (https://maplibre.org/maplibre-gl-js/docs/examples/add-image/).
> 
> I don't know if there is a much bigger risk to do the images loading and layer drawing separately and that it's only a minor and temporary console warning but if we want to follow the documentation example, the image load should probably be inside `setSourcesAndLayers`. What do you think, maybe you already tested that ?

This Maplibre function (`loadImage`) work very wells for small projects where you have static layers with one or two images to load before actually set layers.

In our case, we have dynamic layers and a layer can require many images. Images can be shared between multiples layers

I don't think to display a layer we should:
- Browse the layer config to get all images ids
- Load all these images
- And only when theses images are loaded, set the layer into the map

Technically, I'm confident that having layers where image is not loaded is not a risk.
As you can see, when an image is loaded, the elements of the layer that depend on this image are displayed without any problem.

RafaelSzmarowski (approved):
LGTM ! Thanks for the answer, as far as I have tested it, indeed loading the images after setting the layers doesn't appear like an issue and doesn't crash the map. There could be maybe a small delay in the appearance of the images when loading a lot of images but let's not anticipate too much things that could never happer 🚀

packages/visualizations/src/components/MapPoi/utils.ts
```diff
@@ -26,63 +27,100 @@ export const getMapSources = (sources: PoiMapData['sources']): StyleSpecificatio
...
-// Only circle layers are supported
-export const getMapLayers = (layers?: Layer[]): CircleLayerSpecification[] => {
-    if (!layers) return [];
+const getBaseMapLayerConfiguration = (layer: Layer) => {
+    const { id, source, sourceLayer } = layer;
+    return {
+        id,
+        source,
+        ...(sourceLayer ? { 'source-layer': sourceLayer } : undefined),
+        filter: ['==', ['geometry-type'], 'Point'],
```
etienneburdet, line 36: Seems to work well with `null`
```suggestion
        ...(sourceLayer ? { 'source-layer': sourceLayer } : null),
```

etienneburdet (approved):
LGTM, except for very minor style stuff.

I must say I'm not a huge fan of `icon-image` being either a simple id or an array of death. It makes typing complicated. But well, that's how maplibre works and I lack for a better solution :shrug:

etienneburdet (approved):
LGTM, except for very minor style stuff.

I must say I'm not a huge fan of `icon-image` being either a simple id or a polymorphic array of death. It makes typing complicated. But well, that's how maplibre works and I lack for a better solution :shrug:

etienneburdet:
Just thinking about it: maybe we could add a story with an actual imageMatch to serve as an exemple of what the array could look like.

PR #205, opened 2023-11-17
[FEATURE][SC-44895][POI MAP] Popup improvements

KevinFabre-ods:
## Summary

~~The goal for this PR is to improve popup display on small screen.~~
~~The popup position is forced (take full width) when the map width is below 600px).~~
~~Also add a close icon button on popup (only for sidebar popup or when map size is below 600px)~~

The goal for this PR is to add a new popup display option: Modal
The popup takes the full width available. 
Popup height is based on its content with a `max-height` of `100%`

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-44895](https://app.shortcut.com/opendatasoft/story/44895). Can be tested with https://github.com/opendatasoft/platform/pull/11413
Screenshot 2023-11-27 at 13 36 18

### Changes

- Update Svelte to the last v3 version. ~~Need it to use CSS container queries. Support starts from v3.58.0~~
- Improve accessibility of legend items

## Review checklist

- [x] Description is complete

### DIFF
--- a/packages/visualizations/src/components/MapPoi/Map.ts
+++ b/packages/visualizations/src/components/MapPoi/Map.ts
@@ -10,8 +10,14 @@ import maplibregl, {
     StyleSpecification,
 } from 'maplibre-gl';
 
-import { POPUP_OPTIONS, POPUP_WIDTH } from './constants';
-import type { PopupsConfiguration, CenterZoomOptions } from './types';
+import {
+    POPUP_CLASSNAME,
+    POPUP_DISPLAY_CLASSNAME_MODIFIER,
+    POPUP_OPTIONS,
+    POPUP_WIDTH,
+} from './constants';
+import type { PopupConfigurationByLayers, CenterZoomOptions, PopupDisplayTypes } from './types';
+import { POPUP_DISPLAY } from './types';
 
 const CURSOR = {
     DEFAULT: 'default',
@@ -19,8 +25,22 @@ const CURSOR = {
     DRAG: 'move',
 };
 
+const POPUP_FEATURE_STATE_KEY = 'popup-feature';
+
 type MapFunction = (map: maplibregl.Map) => unknown;
 
+type ActiveFeatureType = MapGeoJSONFeature | null;
+
+function updateFeatureState(
+    map: maplibregl.Map,
+    feature: ActiveFeatureType,
+    stateKey: string,
+    stateValue: unknown
+) {
+    if (!feature) return;
+    const { id, source, sourceLayer } = feature;
+    map.setFeatureState({ id, source, sourceLayer }, { [stateKey]: stateValue });
+}
 export default class MapPOI {
     /** The Map object representing the maplibregl.Map instance. */
     private map: maplibregl.Map | null = null;
@@ -37,20 +57,20 @@ export default class MapPOI {
     /** Array of layer IDs that are not from the base style of the map */
     private layerIds: string[] = [];
 
-    /** Array of source IDs used in the map */
-    private sourceIds: string[] = [];
-
     /** A navigation control for the map. */
     private navigationControl = new maplibregl.NavigationControl({ showCompass: false });
 
     /** A popup for displaying information on the map. */
     private popup = new maplibregl.Popup(POPUP_OPTIONS);
 
-    /** Popups configurations. One configuration by layer */
-    private popupsConfiguration: PopupsConfiguration = {};
+    /** An object to store popup configurations for each layers */
+    private popupConfigurationByLayers: PopupConfigurationByLayers = {};
+
+    /** Value to represent the active display of the popup */
+    private activePopupDisplay: PopupDisplayTypes | null = null;
 
     /** An active GeoJSONFeature. Its information are displayed within the popup. */
-    private activeFeature: MapGeoJSONFeature | null = null;
+    private activeFeature: ActiveFeatureType = null;
 
     /** An array of functions to be executed when the map is ready. */
     private queuedFunctions: Array<MapFunction> = [];
@@ -88,9 +108,7 @@ export default class MapPOI {
             const features = map.queryRenderedFeatures(point, { layers: this.layerIds });
             const isMovingOverFeatureWithPopup =
                 features.length &&
-                features.some((feature) =>
-                    Object.keys(this.popupsConfiguration).includes(feature.layer.id)
-                );
+                features.some((feature) => feature.layer.id in this.popupConfigurationByLayers);
             canvas.style.cursor = isMovingOverFeatureWithPopup ? CURSOR.HOVER : CURSOR.DEFAULT;
         });
     }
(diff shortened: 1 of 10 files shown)


packages/visualizations/src/components/MapPoi/MapRender.svelte
```diff
@@ -124,27 +126,98 @@
...
+        margin-top: 13px;
+        margin-right: 13px;
+    }
+
+    /* --- CONTAINER QUERIES ---  */
+    @container map-card (width < 600px) {
+        /* Force tooltip popup to behave like a sidebar popup */
+        .map-card :global(.poi-map__popup) {
+            height: 100%;
+            max-height: 100%;
```
etienneburdet, line 196: Same here, no reason to have both and seems to work only with height.

packages/visualizations/package.json
```diff
@@ -82,7 +82,7 @@
         "rollup-plugin-svelte": "^6.1.0",
         "rollup-plugin-terser": "^7.0.2",
         "rollup-plugin-visualizer": "^4.2.0",
-        "svelte": "^3.43.2",
+        "svelte": "^3.59.2",
```
etienneburdet, line 85: Not blocking for this PR, but why was it needed? We probably should update to 4 (and 5 quickly) anyways.

etienneburdet (approved):
All good for me! Nice use of container queries :sunglasses:. Comments are minor and pretty optional.

packages/visualizations/package.json
KevinFabre-ods: Need it to use CSS container queries. Support starts from v3.58.0

packages/visualizations/src/components/MapPoi/constants.ts
```diff
@@ -14,8 +14,9 @@ export const DEFAULT_DARK_GREY: Color = '#515457';
 
 export const POPUP_WIDTH = 300;
 
+export const POPUP_CLASSNAME = 'poi-map__popup';
```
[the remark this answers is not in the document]
KevinFabre-ods: `POPUP_CLASSNAME` is used in Map.ts. I'll add a comment next to this variable to explain that if it is edited, styles in `MapRender.svelte` should also be updated.

packages/visualizations/src/components/MapPoi/MapRender.svelte
```diff
@@ -124,27 +126,98 @@
...
-        max-height: calc(100% - 26px);
         height: auto;
-        overflow-y: auto;
+        max-height: 100%;
+        box-sizing: border-box;
     }
     .map-card :global(.poi-map__popup.poi-map__popup--as-sidebar) {
         /* TO DO: add common stylesheet */
-        transform: translate(13px, 13px) !important;
+        height: 100%;
```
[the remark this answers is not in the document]
KevinFabre-ods: I'll check 👍

RafaelSzmarowski (changes requested):
Works very fine and it's a nice improvement !

I only have few remarks more of style and opinions.

I think we shouldn't probably put the 13px padding as it's also the same for + - control panel and that we will always see a shadow behind the tooltip right top corner (please see 
Capture d’écran 2023-11-24 à 11 53 47
pictures)

Capture d’écran 2023-11-24 à 11 43 36

So change one or the other but avoid having this overlapping effect maybe ?

Another thing but it's more opinion based, we shouldn't probably force always the tooltip behaviour to be dependent on the container size, don't think our particular Studio case should impact the SDK component and I find this choice a strong constraint. It should be maybe an option IMO, we can discuss this point IRL if it's easier to make a decision.

Finally, when testing on our use case in Studio it's also strange on small maps to see no difference in the switch between the side panel and classic tooltip but this design issue is out of scope here so we should only focus on what options we want to offer for SDK users.

RafaelSzmarowski:
I think as well that there is a (new ?) warning on a11y for the chart canvas that has a role of img. We can probably ignore that and add a `<!-- svelte-ignore a11y-no-interactive-element-to-noninteractive-role -->` in the file ?

KevinFabre-ods:
> Works very fine and it's a nice improvement !
> 
> I only have few remarks more of style and opinions.
> 
> I think we shouldn't probably put the 13px padding as it's also the same for + - control panel and that we will always see a shadow behind the tooltip right top corner (please see Capture d’écran 2023-11-24 à 11 53 47 pictures)
> 
> Capture d’écran 2023-11-24 à 11 43 36
> 
> So change one or the other but avoid having this overlapping effect maybe ?
> 
> Another thing but it's more opinion based, we shouldn't probably force always the tooltip behaviour to be dependent on the container size, don't think our particular Studio case should impact the SDK component and I find this choice a strong constraint. It should be maybe an option IMO, we can discuss this point IRL if it's easier to make a decision.
> 
> Finally, when testing on our use case in Studio it's also strange on small maps to see no difference in the switch between the side panel and classic tooltip but this design issue is out of scope here so we should only focus on what options we want to offer for SDK users.

In https://github.com/opendatasoft/ods-dataviz-sdk/pull/205/commits/abf917302021da1c5d2ca6f4b1dff5d16b63d8e5 I update `shadow-box` for the popup and navigation controls. 

Screenshot 2023-11-27 at 13 36 18

It's the best I can do without touching other part of the design.

packages/visualizations/src/components/MapPoi/MapRender.svelte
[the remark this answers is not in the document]
KevinFabre-ods: Done in https://github.com/opendatasoft/ods-dataviz-sdk/pull/205/commits/a3c9327339160dffa646a1ebff7f957f78af1fa2

packages/visualizations/src/components/MapPoi/constants.ts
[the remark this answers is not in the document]
KevinFabre-ods: Done in https://github.com/opendatasoft/ods-dataviz-sdk/pull/205/commits/311bae5029aa8287c68faf1d65eb0e6ea13518c9

packages/visualizations/src/components/MapPoi/Map.ts
```diff
@@ -135,13 +141,36 @@ export default class MapPOI {
...
+    /** Update popup display between tooltip, sidebar and modal mode */
+    private updatePopupDisplay(display: PopupDisplayTypes) {
+        // Navigation controls could have been removed if the previous popup display was 'modal'.
+        this.queue((map) => map.addControl(this.navigationControl));
+
+        // Remove all popup display classname modifier before adding the current modifier.
+        Object.keys(POPUP_DISPLAY_CLASSNAME_MODIFIER).forEach((d) => {
+            this.popup.removeClassName(POPUP_DISPLAY_CLASSNAME_MODIFIER[d as PopupDisplayTypes]);
+        });
+        this.popup.addClassName(POPUP_DISPLAY_CLASSNAME_MODIFIER[display]);
```
RafaelSzmarowski, line 153: I have the feeling here that we should know what popup is present and we don't need to really do everytime a loop, isn't it passed through options ?

packages/visualizations/src/components/MapPoi/Map.ts
```diff
@@ -135,13 +141,36 @@ export default class MapPOI {
...
+        /*
+         * Remove navigation controls when popup display is 'modal'.
+         * When the popup closes, add navigation controls to the maps.
+         */
+        if (display === POPUP_DISPLAY.modal) {
+            this.queue((map) => map.removeControl(this.navigationControl));
+            this.popup.once('close', () =>
+                this.queue((map) => map.addControl(this.navigationControl))
+            );
+        }
```
RafaelSzmarowski, line 164: I have a crash in control, I haven't found yet if this is related to this part or somewhere else, do you have it also ? 

Capture d’écran 2023-11-28 à 10 54 08

RafaelSzmarowski (changes requested):
Overall it works and I really like the idea of having different popups but I've found a strange crash in the console and some parts of code that I think are hard to understand. There are many interactions between `setPopupsConfiguration`, `updatePopupDisplay`,  `updatePopup` all this queing and it's seems hard to understand the flow but maybe with few comments and few changes it will be much easier !

packages/visualizations/src/components/MapPoi/Map.ts
```diff
@@ -135,13 +141,36 @@ export default class MapPOI {
 
     private bindedResetLeftPaddingPopup = this.resetLeftPaddingPopup.bind(this);
 
+    /** Update popup display between tooltip, sidebar and modal mode */
+    private updatePopupDisplay(display: PopupDisplayTypes) {
+        // Navigation controls could have been removed if the previous popup display was 'modal'.
+        this.queue((map) => map.addControl(this.navigationControl));
```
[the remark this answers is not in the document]
KevinFabre-ods: This part has been completely refactor in https://github.com/opendatasoft/ods-dataviz-sdk/pull/205/commits/7635413b7ebebc37d6388109f1b126cb5b1426d7

packages/visualizations/src/components/MapPoi/Map.ts
KevinFabre-ods: In https://github.com/opendatasoft/ods-dataviz-sdk/pull/205/commits/7635413b7ebebc37d6388109f1b126cb5b1426d7, add a variable: `activePopupDisplay` to store the current popup display

packages/visualizations/src/components/MapPoi/Map.ts
KevinFabre-ods: Should not happens since in https://github.com/opendatasoft/ods-dataviz-sdk/pull/205/commits/7635413b7ebebc37d6388109f1b126cb5b1426d7

packages/visualizations/src/components/MapPoi/MapRender.svelte
```diff
@@ -134,32 +134,84 @@
...
+
+    /* --- POPUP CONTENT ---  */
     .map-card :global(.poi-map__popup > .maplibregl-popup-content) {
+        display: flex;
+        flex-direction: column;
+        flex-wrap: nowrap;
         padding: 13px;
         border-radius: 6px;
+        max-height: 100%;
+        overflow-y: auto;
```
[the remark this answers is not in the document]
RafaelSzmarowski: Thanks for the fix, it's better indeed, but I still have this in the Studio : 

Capture d’écran 2023-11-30 à 11 06 27

I have the feeling that it comes from the Studio code now : I think it's strange to apply a fixed width on `rods-page-renderer__poi-map__tooltip__inner .image` of `274px` (changing it to 100% seems to fix this), I guess it comes from the fact that the size of the modal popup adapt to the container and will sometimes be smaller than 274px. in the way I see it, you can fix the popup width but the image should adapt and stay centered. It should probably be part of a Studio ticket to improve this and not here 👍

packages/visualizations/src/components/MapPoi/Map.ts
```diff
@@ -122,26 +140,110 @@ export default class MapPOI {
              * We ask for features that are not in base style layers
              */
             const features = map.queryRenderedFeatures(point, { layers: this.layerIds });
-            this.updatePopup(map, features);
+            this.onFeaturesClick(map, features);
```
RafaelSzmarowski, line 143: It's just a slight readability suggestion, open for discussion:

First point, not mandatory at all, maybe rename this function onFeaturesClick to handleFeaturesAndPopups or something more explanatory of what the function do.

Second point, now as this function has the map instance we can probably pass the clicked `point` as an argument and place the` const features = map.queryRenderedFeatures(point, { layers: this.layerIds });` logic inside the function to make the onClick more pure ?

packages/visualizations/src/components/MapPoi/Map.ts
```diff
@@ -313,13 +375,10 @@ export default class MapPOI {
...
-        this.popupsConfiguration = config;
-        this.queue((map) => {
-            if (this.activeFeature) {
-                this.updatePopup(map, [this.activeFeature]);
-            }
-        });
+    setpopupConfigurationByLayers(config: PopupConfigurationByLayers) {
+        this.popupConfigurationByLayers = config;
+        this.updatePopupContent();
+        this.updatePopupDisplay();
```
RafaelSzmarowski, line 381: After our discussions, I understand why we need to update content and display when `PopupConfigurationByLayers` changes but maybe a comment would be useful to understand that `updatePopupContent` and `updatePopupDisplay` use `PopupConfigurationByLayers` so whenever the config changes we must sync it to be sure that the popup has the latest config.

RafaelSzmarowski (approved):
I really like the new way of handling all this popup and features events, IMO it's more readable and maintainable  👍  

I've left few comments, not mandatory, only suggestions ! 

Nice job !

packages/visualizations/src/components/MapPoi/MapRender.svelte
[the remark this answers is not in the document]
KevinFabre-ods: Will check that

packages/visualizations/src/components/MapPoi/Map.ts
KevinFabre-ods: Done in https://github.com/opendatasoft/ods-dataviz-sdk/pull/205/commits/618ec9ed9bc7005cacdea6a652b660ec35318e7a

PR #206, opened 2023-11-20
feat: add cooperative gestures for maps

RafaelSzmarowski:
## Summary

The goal for this PR is to add the option of cooperative gestures for maps, both POI and Choropleth.

This PR needs an update of Maplibre at least up to version 2.2.0 but as the version 2.2.1 fixes some typings issue on Maplibre expressions, it seems better to already update to 2.2.1.

I had troubles to fix the way we defined `FilterExpression`, If you see better solutions, please let me know ! And also this upgrade will have consequences for our users, some types changes will be necessary even if it doesn't appear heavy.

Please check if there is no regression due to the types changes (I've slightly changed the way the functions are made)

There is a change in Chromatic that I can't really explain: https://www.chromatic.com/test?appId=616594e84c1692003af430f3&id=655b8ed6701077f47cbd3365. It doesn't affect many stories so this must be really niche.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-44646](https://app.shortcut.com/opendatasoft/story/44646/studio-poi-maps-choropleth-zoom-on-maps-while-holding-control-command).

### Changes

- Update Maplibre to 2.2.1
- Adapt types changes on Maplibre expressions
- Add some CSS to default cooperative gesture screen rules.

## Open discussion

Do you see better typing changes ?

## To be tested

Regression and Chromatic tests 🙏 

## Review checklist

- [ ] Description is complete

### DIFF
--- a/packages/visualizations/src/components/types.ts
+++ b/packages/visualizations/src/components/types.ts
@@ -39,6 +39,12 @@ export type ColorScale = GradientScale | PaletteScale;
 // We expect an array with couples of values and ExpressionInput and for the last element a default ExpressionInput
 export function isGroupByForMatchExpression(
     value: ExpressionInputType[]
-): value is [ExpressionInputType, ExpressionInputType, ExpressionInputType] {
+): value is [
+    ExpressionInputType,
+    ExpressionInputType,
+    ExpressionInputType,
+    ...ExpressionInputType[],
+    ExpressionInputType
+] {
     return value.length >= 3 && value.length % 2 === 1;
 }
--- a/packages/visualizations/src/components/types.ts
+++ b/packages/visualizations/src/components/types.ts
@@ -1,3 +1,5 @@
+import type { ExpressionInputType } from 'maplibre-gl';
+
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type DataFrame = Record<string, any>[];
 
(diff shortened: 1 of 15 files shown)


KevinFabre-ods (approved):
👌

Minor the lint error 😉

packages/visualizations/src/components/types.ts
```diff
@@ -33,3 +35,10 @@ export type PaletteScale = {
 };
 
 export type ColorScale = GradientScale | PaletteScale;
+
+// We expect an array with couples of values and ExpressionInput and for the last element a default ExpressionInput
+export function isGroupByForMatchExpression(
+    value: ExpressionInputType[]
+): value is [ExpressionInputType, ExpressionInputType, ExpressionInputType] {
+    return value.length >= 3 && value.length % 2 === 1;
+}
```
etienneburdet, line 44: Will tell that it's 3 values or more with an odd number of values (it's straight maplibre.d.ts without the ExpressionSpecification or array of input cases)
```suggestion
export function isGroupByForMatchExpression(
    value: ExpressionInputType[]
): value is [
    ExpressionInputType,
    ExpressionInputType,
    ExpressionInputType,
    ...ExpressionInputType[],
    ExpressionInputType
] {
    return value.length >= 3 && value.length % 2 === 1;
}
```

etienneburdet (approved):
LGTM, I left one minor TS comment that led to think the array was exactly 3 elements, appart from that works super well!

packages/visualizations/src/components/types.ts
RafaelSzmarowski: Thanks ! Made the change: https://github.com/opendatasoft/ods-dataviz-sdk/pull/206/commits/69ca6e33d36510cbf958cc7175ac81cddad02b66.

PR #209, opened 2023-12-05
feat(POI maps): add navigation controls for multi features popup

RafaelSzmarowski:
## Summary

The goal for this PR is ~~to test a simple implementation~~ to add navigation controls inside the popup when the click event query multiple features.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-44648](https://app.shortcut.com/opendatasoft/story/44648/studio-poi-maps-navigate-through-records-in-the-tooltip).

https://github.com/opendatasoft/ods-dataviz-sdk/assets/117300300/6babce90-e045-4f6b-948c-4ea43b883fbf

Native Popup close button has been removed to add a custom arrow which is at the same level of the navigation buttons. 

We also update how a feature is highlight. Highlight and un-highlight strategies are different for each layer type. Currently, only 'symbol' and 'circle' has highlight and un-highlight strategies.

There is a small glitch when a symbol feature is selected: the layer flickers. I didn't find why, but I think this is acceptable.

## Review checklist

- [x] Description is complete

### DIFF
--- a/packages/visualizations-react/stories/Poi/PoiMap.stories.tsx
+++ b/packages/visualizations-react/stories/Poi/PoiMap.stories.tsx
@@ -57,8 +57,20 @@ const layers = [layer1, layer2];
 
 const citiesColorMatch = {
     key: 'key',
-    colors: { Paris: 'blue', Nantes: 'yellow', Bordeaux: 'purple', Marseille: 'lightblue' },
-    borderColors: { Paris: 'white', Nantes: 'black', Bordeaux: 'white', Marseille: 'black' },
+    colors: {
+        Paris: 'blue',
+        'Paris--duplicate': 'lightblue',
+        Nantes: 'yellow',
+        Bordeaux: 'purple',
+        Marseille: 'lightblue',
+    },
+    borderColors: {
+        Paris: 'white',
+        'Paris--duplicate': 'white',
+        Nantes: 'black',
+        Bordeaux: 'white',
+        Marseille: 'black',
+    },
 };
 
 const battleImageMatch = {
--- a/packages/visualizations/src/components/MapPoi/Map.ts
+++ b/packages/visualizations/src/components/MapPoi/Map.ts
@@ -14,6 +14,8 @@ import {
     CONTROL_POSITION,
     POPUP_CLASSNAME,
     POPUP_DISPLAY_CLASSNAME_MODIFIER,
+    POPUP_NAVIGATION_CONTROLS_CLASSNAME,
+    POPUP_NAVIGATION_ARROW_CLASSNAME,
     POPUP_OPTIONS,
     POPUP_WIDTH,
 } from './constants';
@@ -80,6 +82,9 @@ export default class MapPOI {
 
     /** An active GeoJSONFeature. Its information are displayed within the popup. */
     private activeFeature: ActiveFeatureType = null;
+
+    /** All available GeoJSONFeatures on click event */
+    private availableFeaturesOnClick: ActiveFeatureType[] = [];
 
     /** An array of functions to be executed when the map is ready. */
     private queuedFunctions: Array<MapFunction> = [];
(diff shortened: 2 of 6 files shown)


packages/visualizations/src/components/MapPoi/Map.ts
```diff
@@ -259,13 +303,14 @@ export default class MapPOI {
...
             this.popup.remove();
+            this.availableFeaturesOnClick = [];
             return;
         }
 
         // FIXME: remove eslint comment.
         // eslint-disable-next-line prefer-destructuring
         this.activeFeature = features[0];
-
+        this.availableFeaturesOnClick = features;
```
KevinFabre-ods, line 313: As popup is an optional configuration for a layer we must filter `availableFeaturesOnClick` with only feature that has a popup config.

KevinFabre-ods (commented):
Nice ✨ 

I don't see anything that could block this POC.

We should also generate the close button in the same function that generate the arrow buttons

etienneburdet:
I kinda think the popup can be its own Svelte component already, without much modifications to Map.ts actually. If `updatePopupContent` takes a callback, we can use that callback to update a `let availableFeatures`  and then have some `{#if availableFeatures} <Popup {availableFeatures} /> {/if}` and then we can do everything inside the popup component I think.

That would be one less thing to migrate and unless I miss something, not as much work as it seems in the first place.

KevinFabre-ods:
> I kinda think the popup can be its own Svelte component already, without much modifications to Map.ts actually. If `updatePopupContent` takes a callback, we can use that callback to update a `let availableFeatures` and then have some `{#if availableFeatures} <Popup {availableFeatures} /> {/if}` and then we can do everything inside the popup component I think.
> 
> That would be one less thing to migrate and unless I miss something, not as much work as it seems in the first place.

@etienneburdet  If a Svelte component <Popup /> controls the popup instance we must remove the  'popup' variable declaration in the `Map.ts` file. 
This requires to move the `updatePopupContent` and `updatePopupDisplay` functions in the Popup component. Also when a popup is closed, we must [update a map feature state](https://github.com/opendatasoft/ods-dataviz-sdk/blob/d42abd11939ff0b809409d947ab4af72010e6a0d/packages/visualizations/src/components/MapPoi/Map.ts#L504-L506) to change how big a point is. 

IMO all these changes required to completely rework on how the information is circulating between popup and map instance and should be done during [our migration to Svelte components](https://app.shortcut.com/opendatasoft/story/45631/sdk-transform-poi-maps-from-class-to-svelte-component).

packages/visualizations/src/components/MapPoi/Map.ts
KevinFabre-ods: Done here https://github.com/opendatasoft/ods-dataviz-sdk/pull/209/commits/7531261c7c401d9a3b2b90bcd2462ec342dd80d6

KevinFabre-ods:
@RafaelSzmarowski I can't assign you as reviewer but can you please review it. 
Since your work, I updated the code based on the comments we shared.

RafaelSzmarowski:
> @RafaelSzmarowski I can't assign you as reviewer but can you please review it. Since your work, I updated the code based on the comments we shared.

Thanks for the last changes and for taking care of this PR ! As far as I have tested it, it looks very well and the code LGMT 🏅 Nice work !

KevinFabre-ods:
Seen with the design team to always show the navigation controls. So controls must be visible when the content is loading
Done in https://github.com/opendatasoft/ods-dataviz-sdk/pull/209/commits/4610afa7fa6b234fa1b37397f63b9dbef57226d5

https://github.com/opendatasoft/ods-dataviz-sdk/assets/117300300/3267b9cc-3f86-4ea1-ad9d-df534760b4a1

packages/visualizations/src/components/MapPoi/Map.ts
```diff
@@ -30,22 +40,31 @@ const CURSOR = {
     DRAG: 'move',
 };
 
-const POPUP_FEATURE_STATE_KEY = 'popup-feature';
+const ACTIVE_FEATURE_RATIO_SIZE = 1.3;
+
+/** Sorts features in a layer by setting a sort key for a specific feature. */
+const sortLayerFeatures = (
+    map: maplibregl.Map,
```
etienneburdet, line 47: We might as well `import type { Map, …}` I think. We can alias map if need be. It's minor but could lead to think that other types are note from the same namespace.

packages/visualizations-react/stories/Poi/PoiMap.stories.tsx
etienneburdet, line 1: Any reasons for these changes? We might as well minimize the diff.

etienneburdet (approved):
All good for me if we migrate that to a Svelte component later 👌

packages/visualizations-react/stories/Poi/PoiMap.stories.tsx
KevinFabre-ods: Yes of course.
It's to have two records that are on the same coordinates but with different values. 
When switching records in the tooltip, we should see the active feature to be on top of other features for the same coordinates

packages/visualizations/src/components/MapPoi/Map.ts
KevinFabre-ods: done https://github.com/opendatasoft/ods-dataviz-sdk/pull/209/commits/ef13c4ff4f24b9e80597d2a83bde82589d5bd483

PR #211, opened 2023-12-19
feat: add fixed max bounds props

RafaelSzmarowski:
## Summary

The goal for this PR is to add a props to override the default automatic max bounds computation. It seems that with very large bbox as for example the world bbox this automatic computation causes bad rendering depending on the canvas ratio. Overriding it with a new props is not the most satisfying solution but at this point I don't see another solution.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-45683](https://app.shortcut.com/opendatasoft/story/45683/add-world-tileset-in-the-studio-choropleths).

## Review checklist

- [ ] Description is complete

### DIFF
--- a/packages/visualizations/src/components/Map/WebGl/ChoroplethVectorTiles.svelte
+++ b/packages/visualizations/src/components/Map/WebGl/ChoroplethVectorTiles.svelte
@@ -1,5 +1,10 @@
 <script lang="ts">
-    import type { SourceSpecification, GestureOptions, ExpressionSpecification } from 'maplibre-gl';
+    import type {
+        SourceSpecification,
+        GestureOptions,
+        ExpressionSpecification,
+        LngLatBoundsLike,
+    } from 'maplibre-gl';
     import type { BBox } from 'geojson';
     import { debounce } from 'lodash';
     import type { ColorScale, DataBounds, Color, Source } from '../../types';
@@ -48,6 +53,7 @@
     // Data source link
     let sourceLink: Source | undefined;
     let cooperativeGestures: boolean | GestureOptions | undefined;
+    let fixedMaxBounds: LngLatBoundsLike | undefined;
 
     // Used to apply a chosen color for shapes without values (default: #cccccc)
     let emptyValueColor: Color;
@@ -75,6 +81,7 @@
         navigationMaps,
         sourceLink,
         cooperativeGestures,
+        fixedMaxBounds,
     } = options);
 
     // Choropleth is always display over a blank map, for readability purposes
@@ -153,6 +160,7 @@
     {data}
     {sourceLink}
     {cooperativeGestures}
+    {fixedMaxBounds}
 />
 
 <style>
--- a/packages/visualizations/src/components/Map/WebGl/MapRender.svelte
+++ b/packages/visualizations/src/components/Map/WebGl/MapRender.svelte
@@ -69,6 +69,8 @@
     // Data source link
     export let sourceLink: Source | undefined;
     export let cooperativeGestures: boolean | GestureOptions | undefined;
+    // Fixed max bounds that will overide the automatic map.getBounds when setting the bbox
+    export let fixedMaxBounds: LngLatBoundsLike | undefined | null = null;
 
     let clientWidth: number;
     let legendVariant: LegendVariant;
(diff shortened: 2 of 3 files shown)


etienneburdet:
It's a working fix, but to me—and sorry to get back to my old idea :sweat-smile:—the root problem is that we use bounds for zoom, while there to distinct UX variables: where you can zoom/move and what you see. Which is why I've never been a fan of this part of the code (and tried to remove it in old PRs :innocent:):
```
        map.setMaxBounds(null);
        // Using padding, keep enough room for controls (zoom) to make sure they don't hide anything
        map.fitBounds(box as LngLatBoundsLike, {
            animate: false,
            padding: 40,
        });
        // Set new map max bounds after bbox changes
        map.setMaxBounds(fixedMaxBounds || map.getBounds());
```
IMO we should setMaxBounds (or maxZoom) and fitBounds on different props and lifecycle. If we have to se both to the same thing in the Studio, then we can just pass the same thing as props (rather than recomputing).

Now, either we do the refactor here, either we merge this one asap and create a ticket. I'm fine with both.

RafaelSzmarowski:
> It's a working fix, but to me—and sorry to get back to my old idea :sweat-smile:—the root problem is that we use bounds for zoom, while there to distinct UX variables: where you can zoom/move and what you see. Which is why I've never been a fan of this part of the code (and tried to remove it in old PRs 😇):
> 
> ```
>         map.setMaxBounds(null);
>         // Using padding, keep enough room for controls (zoom) to make sure they don't hide anything
>         map.fitBounds(box as LngLatBoundsLike, {
>             animate: false,
>             padding: 40,
>         });
>         // Set new map max bounds after bbox changes
>         map.setMaxBounds(fixedMaxBounds || map.getBounds());
> ```
> 
> IMO we should setMaxBounds (or maxZoom) and fitBounds on different props and lifecycle. If we have to se both to the same thing in the Studio, then we can just pass the same thing as props (rather than recomputing).
> 
> Now, either we do the refactor here, either we merge this one asap and create a ticket. I'm fine with both.

I don't really understand how a maxZoom could resolve the world bbox and max bounds problem, but I'm probably missing something, I guess that if you're ok with that, we can apply this fix for our particular problem, merge this and create a separate ticket to try to solve all our bbox, max bounds problems, it would be maybe more clear and the feature on Studio side won't wait for too long.

RafaelSzmarowski:
> > It's a working fix, but to me—and sorry to get back to my old idea :sweat-smile:—the root problem is that we use bounds for zoom, while there to distinct UX variables: where you can zoom/move and what you see. Which is why I've never been a fan of this part of the code (and tried to remove it in old PRs 😇):
> > ```
> >         map.setMaxBounds(null);
> >         // Using padding, keep enough room for controls (zoom) to make sure they don't hide anything
> >         map.fitBounds(box as LngLatBoundsLike, {
> >             animate: false,
> >             padding: 40,
> >         });
> >         // Set new map max bounds after bbox changes
> >         map.setMaxBounds(fixedMaxBounds || map.getBounds());
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > IMO we should setMaxBounds (or maxZoom) and fitBounds on different props and lifecycle. If we have to se both to the same thing in the Studio, then we can just pass the same thing as props (rather than recomputing).
> > Now, either we do the refactor here, either we merge this one asap and create a ticket. I'm fine with both.
> 
> I don't really understand how a maxZoom could resolve the world bbox and max bounds problem, but I'm probably missing something, I guess that if you're ok with that, we can apply this fix for our particular problem, merge this and create a separate ticket to try to solve all our bbox, max bounds problems, it would be maybe more clear and the feature on Studio side won't wait for too long.

To give a little context on my issue: the bbox is fixed and given for a specific context whereas the maxBounds must adapt to a specific map ratio display, that's why we first apply the bbox, then fit the map and only after we give the proper maxBounds. I really must be missing some logic that you see because I don't see a quick fix using maxZoom. I really think that reworking this logic should be separate in another bigger ticket as it doesn't appear trivial.

etienneburdet:
Let's merge this as is and see how we split the two in https://app.shortcut.com/opendatasoft/story/46659

etienneburdet (approved):
Let's merge it as is.

KevinFabre-ods (approved):
LGTM

Let's keep the technical discussion in the story that you've created Etienne.

PR #221, opened 2024-01-26
fix(POI Maps): prevent change in sort order for symbol features

KevinFabre-ods:
## Summary

The goal for this PR is to prevent features in symbol layer to change of sort order. This result in a visual glitch where symbol feature change sort order when a feature is active (clicked)

Once we set a expression for the 'symbol-sort-key' property we need to reset the same expression. Otherwise sort order will change once the popup is removed. 

## To be tested

Link to the platform and create POI Maps with a large number of symbol

## Review checklist

- [x] Description is complete


etienneburdet (approved):
It works if wee need a quick fix 👌 

Big scheme of things, I think we should use `Marker` for this, but it can wait.

PR #223, opened 2024-02-12
[FEAT][TABLE][SC-46247] Basic version

KevinFabre-ods:
## Summary

The goal for this PR is to add a simple version of a table component. This take inspiration from the exploration: https://github.com/opendatasoft/ods-dataviz-sdk/pull/222

(Internal for Opendatasoft only) Associated Shortcut tickets: 
- [sc-46247](https://app.shortcut.com/opendatasoft/story/46247)
- [sc-46255](https://app.shortcut.com/opendatasoft/story/46255)

### Changes

- Render and style a basic table
- Add visualization common properties: title, subtitle and source link
- Add Two Storybook stories, default and a custom theme
- Make it available as a component in the `visualizations/react` package

## Review checklist

- [x] Description is complete


KevinFabre-ods:
@etienneburdet To let you know that we've started the development with a basic version

packages/visualizations/src/components/Table/Table.svelte
```diff
@@ -0,0 +1,103 @@
...
+        background-color: var(--header-background-color);
+    }
+    :global(.container thead) {
+        border-bottom: 2px solid var(--header-border-color);
+    }
+    :global(.container tbody tr:not(:last-child)) {
+        border-bottom: 1px solid var(--border-color);
+    }
+    :global(.container tbody tr:hover) {
+        background-color: var(--active-row-background-color);
```
etienneburdet, line 97: I think this could be scoped into each component, if we are to use css var theming.

packages/visualizations/src/components/Table/Table.svelte
```diff
@@ -0,0 +1,103 @@
...
+<div class="container" {style}>
+    <div class="header">
+        {#if title}
+            <h3>{title}</h3>
+        {/if}
+        {#if subtitle}
+            <p>{subtitle}</p>
+        {/if}
+    </div>
+    <div class="table-wrapper">
```
etienneburdet, line 34: Could be applied directly on `<table>` with `border-collapse: seperate;`

etienneburdet (approved):
LGTM functionnally. 

Implem wise I'm not a fan of JSON/props based themed, since it makes a lot of discrete options to document, read, pass down as global… I'd rather style directly in CSS. But that's debatable :shrug:

etienneburdet (commented):
Ha yes, also the theme persists when switching stories, but I think you fix that in the style PR

etienneburdet (commented):
Ha yes, also the theme persists when switching stories, but I think you fix that in the style PR

packages/visualizations/src/components/Table/Table.svelte
KevinFabre-ods: Yes, I moved these style in https://github.com/opendatasoft/ods-dataviz-sdk/pull/224

packages/visualizations/src/components/Table/Table.svelte
KevinFabre-ods: `.table-wrapper` is to control the scroll behavior. 
I'll try to use the border styles only the table tag

KevinFabre-ods:
> Ha yes, also the theme persists when switching stories, but I think you fix that in the style PR

Yes, I fixed that with next PRs

packages/visualizations/src/components/Table/utils/theme.ts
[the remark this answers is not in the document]
KevinFabre-ods: It's true that if we exposed an `unstyled`  prop it is in conflict with the theme prop.
https://github.com/opendatasoft/ods-dataviz-sdk/pull/224 

So you are suggesting to remove the `theme` prop and let developers style the table via CSS directly on table tags (like th, td ...) ?

packages/visualizations/src/components/Table/Table.svelte
etienneburdet: Ho, I see. Maybe `display: block; overflow-x: auto` then (I didn't test). But the would avoid the small blank space between border and table when scrolling to far (it comes back, but can "stretch" a bit too far).

packages/visualizations/src/components/Table/utils/theme.ts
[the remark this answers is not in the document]
etienneburdet: I would think so, yes. Unless we maybe want to expose a very basic version with 2-5 colors (that would fit the Studio well indeed).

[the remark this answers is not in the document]
KevinFabre-ods: The table will takes colors from the domain theme.
I'll set neutral colors as default for the SDK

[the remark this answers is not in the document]
KevinFabre-ods: It is something like that you have in mind ? 
https://github.com/opendatasoft/ods-dataviz-sdk/pull/223/commits/526680781e883116de1d825e524f8965b319ef5a

packages/visualizations/src/components/Table/Table.svelte
KevinFabre-ods: I found  `overscroll-behavior-x: none;`to prevent the 'stretch'

KevinFabre-ods:
@etienneburdet I edit a bit the code. Can you check if it's good to start as it is ?

etienneburdet:
All good for me 🏂

PR #226, opened 2024-02-14
[FEAT][TABLE] Add locale options

KevinFabre-ods:
## Summary

The goal for this PR is to add a locale prop to format numbers and dates

### Changes

- New optionnal prop for the table: `locale`
- Add a context at the root of the table component. It's use a writable to store the current locale. Default value comes from `navigator.language`.

## Review checklist

- [x] Description is complete

### DIFF
--- a/packages/visualizations/src/components/Table/Cell/Format/DateFormat.svelte
+++ b/packages/visualizations/src/components/Table/Cell/Format/DateFormat.svelte
@@ -1,22 +1,24 @@
 <script lang="ts">
+    import store from '../../store';
     import type { DateColumn } from '../../types';
+
     import { warn } from './utils';
 
     export let rawValue: unknown;
     export let options: DateColumn['options'];
 
-    function getDisplayValue(v: unknown) {
+    function getDisplayValue(v: unknown, locale: string) {
         if (
             (typeof v === 'string' || typeof v === 'number') &&
             !Number.isNaN(new Date(v).getTime())
         ) {
-            return new Intl.DateTimeFormat(navigator.language, options).format(new Date(v));
+            return new Intl.DateTimeFormat(locale, options).format(new Date(v));
         }
         warn(v, 'date');
         return v;
     }
 
-    $: value = getDisplayValue(rawValue);
+    $: value = getDisplayValue(rawValue, $store.locale);
 </script>
 
 {value}
--- a/packages/visualizations/src/components/Table/store.ts
+++ b/packages/visualizations/src/components/Table/store.ts
@@ -1,0 +1,17 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { writable } from 'svelte/store';
+
+type TableStore = {
+    locale: string;
+};
+
+const defaultLocale = navigator.language;
+
+const store = writable<TableStore>({ locale: defaultLocale });
+
+const updateLocale = (locale: string = defaultLocale) => {
+    store.update((previousValue) => ({ ...previousValue, locale }));
+};
+
+export { updateLocale };
+export default store;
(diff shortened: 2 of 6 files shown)


packages/visualizations/src/components/Table/Cell/Format/DateFormat.svelte
```diff
@@ -1,22 +1,26 @@
...
-    function getDisplayValue(v: unknown) {
+    const context = getTableContext();
+
+    function getDisplayValue(v: unknown, locale: string) {
         if (
             (typeof v === 'string' || typeof v === 'number') &&
             !Number.isNaN(new Date(v).getTime())
         ) {
-            return new Intl.DateTimeFormat(navigator.language, options).format(new Date(v));
+            return new Intl.DateTimeFormat(locale, options).format(new Date(v));
```
etienneburdet, line 17: Could default to navigator.language? Not useful for us—almost a bit dangerous if we forget to set it somehow—but could be nice.

packages/visualizations/src/components/Table/context.ts
etienneburdet, line 1: We don't need to use both a context and a store here. If we need a store, we can import the store directly in the component.

But I'm not even sure we need a store. If `locale` is a prop in the first place, it should be reactive.

etienneburdet (changes requested):
My main request for change would be to use either context or store, or none, but not both. A store is accessible from anywhere it's imported so it doesn't have to be scoped in a context.

My preferred way though would be to just pass local as props. It's a prop of `Table`, it adds a bit of props drilling but it's reasonable for now I think.

packages/visualizations/src/components/Table/Cell/Format/DateFormat.svelte
KevinFabre-ods: Currently with this PR the default locale is `navigator.language` (see context.ts). 
Indeed for our use case, we'll always give a `locale` prop. In we don't setup navigator.language as default locale, which value should we set ? a hardcoded locale 'en' ?

packages/visualizations/src/components/Table/context.ts
KevinFabre-ods: A store is shared between components. Not an issue with the react version of table but what if we have developers that use many Svelte Table components in their Svelte app ?
Table will share values in the store ? 

As context is not reactive it's why I used a store.

I've got a bit ahead and I assume that later on we'll need to centralise a lot of information to avoid a lot of prop drilling (sorting status, pagination).

packages/visualizations/src/components/Table/Cell/Format/DateFormat.svelte
etienneburdet: Yes, I missed the context part early in this review, indeed, sorry! It's alright if it defaults to `navigator.language`👌

packages/visualizations/src/components/Table/context.ts
etienneburdet: For now I think we can assume the `.svelte` will encapsulate the store as needed and is an "implementation detail" for a Svelte dev (we don't even package svelte files actually…). 

But in fine, we could expose the store as part of the API. A svelte store actually is Vanilla JS, so it can be imported and susbscribed anywhere (the `$` is just sugar syntax). For the record we explored that possibility for the filters at some point.

Back to the point: I think should try an implem with just the store, not the context and see if that works well for us now.

KevinFabre-ods: So I updated this part to only use a store

packages/visualizations/src/components/Table/store.ts
```diff
@@ -0,0 +1,17 @@
...
+    locale: string;
+};
+
+const defaultLocale = navigator.language;
+
+const store = writable<TableStore>({ locale: defaultLocale });
+
+const updateLocale = (locale: string = defaultLocale) => {
+    store.update((previousValue) => ({ ...previousValue, locale }));
+};
```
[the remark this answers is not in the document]
KevinFabre-ods: For now let's keep it like this. 

And feel free to break this store when developing the sort feature 😉

PR #230, opened 2024-03-21
Chore/simplify-table-styling

etienneburdet:
## Summary

The goal for this PR is to simplify default styling in the table component. 

The proposed formant can serve as a basis for all component cards. If we like it we can transpose that.

I left out cell formatting for now, to have it in a dedicated story (accounting for custom foramtting).

### Changes
I removed the `index.scss` so that we can have styles where they are used (more common in Svelte) and scope mor things.

I added a `<Card>` component, that is responsible for titlbe, source etc. but also applying default style. Hopefully, this is useable on other components.


KevinFabre-ods (commented):
I'm in favor of this PR, meaning placing each style as closely as possible to the component. Adding the Card component is a nice addition. Now, we just need to use it elsewhere ;p

Be careful not to make the pagination PR harder to read if we merge this code into it. Anyway, we have Chromatic to be the watchdog.

etienneburdet:
Merged into #229

PR #232, opened 2024-04-05
Feature/add-number-per-pages

etienneburdet:
## Summary

The goal for this PR is to add page size options to pagination.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-46253](https://app.shortcut.com/opendatasoft/story/46253). (same as base pr

### Changes
The biggest change is that I removed all kind of internal state for the current page or the current number per page. This proves impractical when you add number per pages, leading to having to track the current page in the app on top of the SDK component itself. So now the current page and the current number per page are fully controlled and determined by a single prop.

The select is a native select too. We currently don't support children/slots, so it's possible to use a platform select or something like that. But a native select should be fine for the use.

Options have both a label and a value because we may want to add extra wording (e.g. "5 / page") or other labelling ("small", "large" etc.)

## Review checklist

- [ ] Description is complete


packages/api-client/package-lock.json
KevinFabre-ods, line 1: We should have change in these package-lock.json files

packages/visualizations-react/stories/Table/custom-style.css
KevinFabre-ods, line 1: Any reason to remove this style ? 
The Table > Custom story has a diff in snapshot

KevinFabre-ods (commented):
At first sight, everything looks good.

Expect a diff in a Chromatic snapshot that, IMO, should not be there. 

I'll let @cybervinvin do the functional review as he implementing the table in the platform

packages/visualizations/src/components/Pagination/PageSize.svelte
```diff
@@ -0,0 +1,33 @@
...
+    $: if (value !== selected) {
+        value = selected;
+    }
+</script>
+
+<select bind:value on:change={(e) => onChange(toNumber(e.currentTarget.value))}>
+    {#each options as option}
+        <option label={option.label} value={option.value} />
+    {/each}
+</select>
```
KevinFabre-ods, line 24:
```suggestion
    export let recordsPerPage: number;
</script>

<select value={recordsPerPage} on:change={(e) => onChange(toNumber(e.currentTarget.value))}>
    {#each options as option}
        <option label={option.label} value={option.value} />
    {/each}
</select>
```

Why not stick to the same pattern that we use to react on a page change ?

packages/visualizations-react/stories/Table/PaginatedTemplates.tsx
```diff
@@ -0,0 +1,132 @@
...
+    return <Table data={paginatedData} options={stateFulOptions} />;
+};
+
+/* This template will fail to catch a new page and returns previous  data: {
+      value,
+      loading: false,
+  }, page and pageSize
+simulating e.g. an API call fail.
+The select should stay on it's previous value, not the clicked one.
+*/
```
KevinFabre-ods, line 109: IMO, we should build an optimistic component that doesn't handle case of failure. 
It up to the parent that use the table to prevent invalid prop configuration.

KevinFabre-ods (commented):
Works fine. 

Maybe we could take some times to discuss how each other see if the table should prevent fail behaviors from its events

packages/visualizations-react/stories/Table/PaginatedTemplates.tsx
etienneburdet: It's a bit different in the case of a native select: if the user selects a value, the value will be set in the HTML with no pass into the JS at all. So if the user clicks on '10' but say the API fail etc. all the Svelte app will remain the same (say on 5), but the select will still display 10.

KevinFabre-ods: OK I get it, that explain: 
```
    $: if (value !== selected) {
        value = selected;
    }
```

packages/visualizations/src/components/Pagination/PageSize.svelte
etienneburdet: It won't work with an HTML select that has it own state. We need a way to map what is in the select to our JS state and then force it back to the select.

KevinFabre-ods: Resolved see https://github.com/opendatasoft/ods-dataviz-sdk/pull/232#discussion_r1601843599

KevinFabre-ods (approved):
Let's merge!

PR #233, opened 2024-04-22
Table: onClick header and sort stories

etienneburdet:
## Summary

The goal for this PR is to add a sort system, by making the header clickable and adding a sort direction indicator

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-48046](https://app.shortcut.com/opendatasoft/story/48046).

### Changes
I simply added a `on:click` and an icon in the header. Same as for pagination, we way want to deal with api calls so it doesn't really make sense to implement sort function. The UI is then just a controlled/stateless interface.

Sorties kinda exemplify what the result could be (sort function of the "api" is stupid on only really works on numbers).

### Changelog
Table now have a sort mechanism.

## Open discussion
The current implementation lacks definitive icons. But maybe we could accept any icon as prop? This opens a lot of questions however (security with SVG, format compitibility etc.), but might not be that hard.

## Review checklist

- [x] Description is complete


packages/visualizations/src/components/Table/types.ts
```diff
@@ -8,9 +8,17 @@ export type DataFormat = typeof DATA_FORMAT[DataFormatKey];
...
+    asc: 'ASC',
+    desc: 'DESC',
+} as const;
+export type ColumnSortValues = typeof ColumnSort[keyof typeof ColumnSort];
+
 type BaseColumn = {
     key: string;
     title: string;
+    sorted?: ColumnSortValues;
+    onClick?: () => void;
```
KevinFabre-ods, line 21: If we have to handle mutli sort, how this could evolve ? 
Add another property like sort position or update `sorted` to a tuple like `['ASC', 0]` ? 

Maybe we could benefit to move the sort pros as a first level `options` prop where it's an array (to handle multi sort) where we store an object with the field key and sort order


```js
sort: [
   {key: 'fieldA', sort: 'ASC'},
   {key: 'fieldB', sort: 'DESC'},
   // ...
]
```

packages/visualizations/src/components/Table/types.ts
```diff
@@ -8,9 +8,17 @@ export type DataFormat = typeof DATA_FORMAT[DataFormatKey];
...
+export const ColumnSort = {
+    asc: 'ASC',
+    desc: 'DESC',
+} as const;
+export type ColumnSortValues = typeof ColumnSort[keyof typeof ColumnSort];
+
 type BaseColumn = {
     key: string;
     title: string;
+    sorted?: ColumnSortValues;
```
KevinFabre-ods, line 20: I would add a comment to explain what kind of value is expected. 

As `ColumnSortValues` is computed from other types, it's might not be easy to understand what is expected

packages/visualizations/src/components/Table/Header.svelte
```diff
@@ -1,14 +1,20 @@
...
+    import { ColumnSort, type Column } from './types';
 
     export let columns: Column[];
 </script>
 
 <thead>
     <tr>
         {#each columns as column}
-            <th class={`table-header--${column.dataFormat}`}>
+            <th class={`table-header--${column.dataFormat}`} on:click={column?.onClick}>
```
KevinFabre-ods, line 10: `th` nodes are not accessible via keyboard navigation (at least I couldn't during my tests, maybe there is keyboard shortcut that I don't know), we can't trigger a click from them when pressing enter. 

Even if we have no design for the sort system, I think we should consider adding this mechanism from the start

KevinFabre-ods (commented):
Personally, I can see UI and UX improvements, but I won't ask to do it in this PR.
Because, at some point, the design team will ask to change it based on a design (hopefully).

So I focused my review only on the technical level

packages/visualizations/src/components/Table/Header.svelte
cybervinvin: A good example: https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/

packages/visualizations/src/components/Table/types.ts
etienneburdet: As it is implemented, it's not a "sort" function, it's just a visual indicator that this column is sorted. What the outter app decides to do with it isn't SDK's problem imo. But whatever they do, the visual indication will remain the same I think.

packages/visualizations/src/components/Table/types.ts
etienneburdet: Intellisense gives the right possible values, no? We never really committed to jsdoc or anything for TS files as far as I know and pretty much rely on intellisense. I can add an obvious comment 🤷‍♂️

KevinFabre-ods:
We now have design for the headers 🎊 
https://app.zeplin.io/project/662a9454cdce0eb321bef34d/screen/664fad0242a557cce9bfb0e2

packages/visualizations/src/components/Table/types.ts
KevinFabre-ods: Yes, I agree we should keep the prop as a visual indicator on how the outer app has decided to represent data.

My point was to merge the sorted key for each column into a single prop at the root of the options prop to ease the future dev if we need to display a sort order indication between fields.

packages/visualizations/src/components/Table/types.ts
KevinFabre-ods: Indeed Intellisense gives the values. I think I just hovered sorted during my testing.
I can confirm that with Intellisense `ColumnSortValues` we got the values

packages/visualizations/src/components/Table/Headers/SortButton.svelte
```diff
@@ -0,0 +1,28 @@
...
+    import type { ColumnSortValues } from '../types';
+    import SortIcon from './SortIcon.svelte';
+
+    export let sorted: ColumnSortValues | undefined;
+</script>
+
+<button type="button" on:click>
+    <slot />
+    <SortIcon {sorted} />
+</button>
```
KevinFabre-ods, line 11: I would also add an aria-label that can be customized with an optional `options` prop

Default values could be: Sort by ascending/descending order

KevinFabre-ods (commented):
Looks good ! 

Just a comment about aria label for the sort button

Also, what your opinion about the answer I gave you for this discussion https://github.com/opendatasoft/ods-dataviz-sdk/pull/233#discussion_r1611296107 ?

packages/visualizations/src/components/Table/types.ts
etienneburdet: Both could work I guess, but I'd rather keep columns enclosing and add a `sortPrority` key later, if we ever need it (which I doubt). The only occurence of a multisort I know of is airtable and the sort order is not in the table itself.

KevinFabre-ods: Ok let's see how the product will drive this feature in the future

PR #236, opened 2024-05-13
[sc-46254] Add CSS style for vertical scroll

cybervinvin:
## Summary

The goal for this PR is to add a fixed header when the `Table` is scrollable vertically.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-46254](https://app.shortcut.com/opendatasoft/story/46254/sdk-table-fixed-headers-and-footers).

https://github.com/opendatasoft/ods-dataviz-sdk/assets/78358994/e07eb056-d3fc-4fde-a65f-5fd0dd8e8050

### Changes

- add `verticalScroll` property
- add new CSS entry `header-background-color`
- add new CSS entry `vertical-scroll-height`
- `thead` is sticky on top
- new story for `Vertical Scroll`

#### Breaking Changes
None

### Changelog

In the `Table` component, the header is now sticky on top when the table is scrollable vertically.

## Open discussion

When the `verticalScroll` property is set to `true`, table height is set arbitrarily and the background color too (default: #ffffff). It's not ideal but I don't see an alternative.

## To be tested

A new story has been created for storybook.

## Review checklist

- [X] Description is complete

### DIFF
--- a/packages/visualizations-react/stories/Table/Table.stories.tsx
+++ b/packages/visualizations-react/stories/Table/Table.stories.tsx
@@ -28,6 +28,12 @@ Playground.args = {
     options,
 };
 
+export const VerticalScroll = Template.bind({});
+VerticalScroll.args = {
+    data,
+    options: { ...options, verticalScroll: true },
+}
+
 const CustomStyleTemplate: ComponentStory<typeof Table> = args => (
     <div className="table-story--custom-style">
         <Table {...args} />
--- a/packages/visualizations-react/stories/Table/Table.stories.tsx
+++ b/packages/visualizations-react/stories/Table/Table.stories.tsx
@@ -28,11 +28,17 @@ Playground.args = {
     options,
 };
 
-export const VerticalScroll = Template.bind({});
+const BoxSizingTemplate: ComponentStory<typeof Table> = args => (
+    <div className='table-story--box-sizing'>
+        <Table {...args} />
+    </div>
+);
+
+export const VerticalScroll = BoxSizingTemplate.bind({});
 VerticalScroll.args = {
     data,
     options: { ...options, verticalScroll: true },
-}
+};
 
 const CustomStyleTemplate: ComponentStory<typeof Table> = args => (
     <div className="table-story--custom-style">
--- a/packages/visualizations/src/components/Table/Table.svelte
+++ b/packages/visualizations/src/components/Table/Table.svelte
@@ -15,7 +15,7 @@
 
 <div class:vscroll-wrapper={verticalScroll}>
     <table aria-describedby={description ? tableId : undefined}>
-        <Headers {columns} {verticalScroll}/>
+        <Headers {columns} {verticalScroll} />
         {#if records}
             <Body {records} {columns} />
         {/if}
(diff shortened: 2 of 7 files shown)


packages/visualizations/src/components/Table/Table.svelte
```diff
@@ -43,4 +44,10 @@
         white-space: nowrap;
         width: inherit;
     }
+
+    .vscroll-wrapper {
+        display: inline-block;
+        height: var(--vertical-scroll-height);
```
etienneburdet, line 50: I'd rather have that as 100% and have the container in the app fix the height. Case is, in the Studio we often define height by ratio.

This implies adding an `height: 100%; flex-wrap: no wrap; ` on the Card.

packages/visualizations/src/components/utils/Card.svelte
```diff
@@ -47,6 +47,8 @@
         --spacing-75: 9px;
         --spacing-100: 13px;
         --border-color: #cbd2db;
+        --header-background-color: #ffffff;
+        --vertical-scroll-height: 20em;
         /* FIXME: Only using flex style to center source link */
         display: flex;
```
etienneburdet, line 53: Could a use a `flex-wrap: nowrap`. It's a direct consequence of this PR, but it becomes more obvious now. Else we can end up in horizontal layout in some cases.

packages/visualizations-react/stories/Table/Table.stories.tsx
```diff
@@ -28,6 +28,12 @@ Playground.args = {
     options,
 };
 
+export const VerticalScroll = Template.bind({});
```
etienneburdet, line 31: So if we go for the 100% on the Table, we'll need to add a container with a fixed height (and maybe one with an `aspect-ratio`?) to match our use cases.

etienneburdet (changes requested):

etienneburdet (changes requested):
Just one thing about height, that I would prefer to have set by the outer app, notably for our aspect ratio use case.

Else I'm fine with all the implem: sticky, conditional class, props to header 👌

packages/visualizations/src/components/Table/Table.svelte
cybervinvin: done

packages/visualizations/src/components/utils/Card.svelte
cybervinvin: done

packages/visualizations-react/stories/Table/Table.stories.tsx
cybervinvin: The container was added to the story but I did not create a new story with`aspect-ratio` (I did not know the property)

KevinFabre-ods:
@cybervinvin with the latest merges, we now have conflicts in this PR (so sorry 🥹 ) 

No rush to fix it. Please stay focus on your dev (loading state) and your reviews

cybervinvin:
No problem, I'll fix it 😉

KevinFabre-ods:
We will re-open it when we are asked to support it.

PR #239, opened 2024-05-22
feat: add callback formatting functinos

etienneburdet:
## Summary

The goal for this PR is to add formatting callbacks to cell, to allow for more flexible renders

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-50243](https://app.shortcut.com/opendatasoft/story/50243).

### Changes
I mainly added callbacks and their default into the options of each of the formatting component. I added on top of the format components and `Intl` options, because I think they bring different things (HTML tags, i18n, console warns etc.). If a dev really needs to fully customize say a date, then they could always use a text column with a custom callback.

For ease of use, I expanded the props of each the components (avoids reactive destructuring) and spread at the top level. I also renamed `getDisplayValue` to `format` in each component to avoid the too similar `display` and `displayValue`.

I also simplified the locale store in a fashion that would add other stores if needed, rather than keys into a big one, since Svelte stores are more made to be many small stores.

I try to add formatting to all default cases, to display different possibilities (e.g. for units in number, one with intl, one with callback).

#### Breaking Changes
`options: { } ` props to columns are a bit different now, stories should be updated accordingly.

### Changelog
The Table columns now accept custom formatting callbacks.

## Open discussion
I don't think we can rely on units only: we have no guarantees they are supported and we may need more (ranks as in "#1", capitalize etc.).

## Review checklist

- [x] Description is complete


packages/visualizations/src/components/Table/types.ts
```diff
@@ -30,7 +34,10 @@ export type NumberColumn = BaseColumn & {
      *
      * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#options
      */
-    options?: Intl.NumberFormatOptions;
+    options?: {
+        display?: (v: string) => string;
+        intl?: Intl.NumberFormatOptions;
```
KevinFabre-ods, line 39: Please move the comment about `NumberFormat` options just above the intl property

packages/visualizations/src/components/Table/types.ts
```diff
@@ -40,12 +47,17 @@ export type DateColumn = BaseColumn & {
      *
      * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat#using_options
      */
-    options?: Intl.DateTimeFormatOptions;
+    options?: {
+        display?: (v: string) => string;
+        intl?: Intl.DateTimeFormatOptions;
```
KevinFabre-ods, line 52: Same thing for the `DateTimeFormat` option comment

packages/visualizations/src/components/Table/store.ts
```diff
@@ -1,17 +1,7 @@
...
-const store = writable<TableStore>({ locale: defaultLocale });
-
-const updateLocale = (locale: string = defaultLocale) => {
-    store.update((previousValue) => ({ ...previousValue, locale }));
-};
-
-export { updateLocale };
-export default store;
+// eslint-disable-next-line import/prefer-default-export
+export const locale = writable<string>(defaultLocale);
```
KevinFabre-ods, line 7: The `locale` prop doesn't seems to be reactive anymore. 

I can update once the locale and see the change, but after that, any change of value for the locale no longer change the format (dates for example)

You can check on the Storybook of the main branch to see the difference with this branch 

Adding a react test, might prevent us to break this in future update. If I remember well, it's the second time we encountered this issue.

packages/visualizations/src/components/Table/types.ts
KevinFabre-ods, line 39: We might want to explain with a comment how this two props works if both defined
1. Intl is done
2. display function run after

KevinFabre-ods (changes requested):
Well done 👏 , I functionally tested it when build column options for the platform repo

Two comments must be moved and I think we have an issue with the reactivity of the `locale` prop

packages/visualizations/src/components/Table/store.ts
etienneburdet: Ok, I looked into it and it's not trivial at all and not directly related to the store. The $locale is still reactive, the problem arrise when we call `display` (even with a fixed locale actually). I suspect this is because we have nested functions, one being a prop, but I can't seem to find my way around it for now… 🤔

KevinFabre-ods: Oh ok, so it might not be related to Svelte but Storybook. 
When some props are functions, Storybook stories might be buggy if we update some props

Could we try to add React tests to asset that both locale and display work as expected ?

etienneburdet:
Sooooo… sorry for the delay, but I wanted to test as much as I could. Here it is: 
* Upgrading everything Storybook  8 and Svelte 4, gives better error and a bit more stability, that is if you hot reload once, then everything works (🤷 ).
* Everything else in storybook works if we don't call `display` : locale update, display, reactive formatting etc.

I didn't include the whole update in this PR, but since I did it, I will open another PR (it's quite big).

packages/visualizations/src/components/Table/store.ts
KevinFabre-ods: So I just found the reason behind our bug

The display function is not a dependency for the computed `value` as we did it for the locale prop.

So when `display` changes, value doesn't change. 

We should either pass display as an argument of the format function or declare the `format` function via a computation

etienneburdet:
All good for me, we can ship like this. Derived store hasn't much use in fact, since we don't want to export `format` or `value` outside of the component. Can merge on my side 🚀

KevinFabre-ods:
@cybervinvin Do you want to do a round of review before we merge this PR?

KevinFabre-ods (approved):
🚀

cybervinvin:
Sorry for lag, it was good for me (I received responses to my comments during our meeting this morning).

PR #240, opened 2024-05-29
Loading records implementation

cybervinvin:
## Summary

The goal for this PR is to add a loading indicator in the rows, when `Table` awaiting new data. 

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-48032](https://app.shortcut.com/opendatasoft/story/48032/sdk-studio-implement-loading-records).

### Changes

Add new subcomponent `LoadingCell` who is displayed when the property `isLoading` is true.

#### Breaking Changes

None

### Changelog

In the Table component a new svg is displayed when fetching the data.

## Open discussion

There are 2 issues that I cannot resolve:
- the SVG is too high in the cell and I don't understand why
- the SVG is not displayed correctly in the different table cells (it only seems to fit the first cell)

## To be tested

New story in storybook

## Review checklist

- [X] Description is complete

### DIFF
--- a/packages/visualizations-react/stories/Table/Table.stories.tsx
+++ b/packages/visualizations-react/stories/Table/Table.stories.tsx
@@ -20,6 +20,11 @@ const data: Async<TableData> = {
     loading: false,
 };
 
+const fetchingData: Async<TableData> = {
+    value: [],
+    loading: true,
+};
+
 const Template: ComponentStory<typeof Table> = args => <Table {...args} />;
 
 export const Playground = Template.bind({});
@@ -76,4 +81,12 @@ export const Unstyled = Template.bind({});
 Unstyled.args = {
     data,
     options: { ...options, unstyled: true },
+};
+
+export const Loading = Template.bind({});
+Loading.args = {
+    data: fetchingData,
+    options: {
+        ...options,
+    },
 };
--- a/packages/visualizations-react/stories/Table/Table.stories.tsx
+++ b/packages/visualizations-react/stories/Table/Table.stories.tsx
@@ -86,7 +86,5 @@ Unstyled.args = {
 export const Loading = Template.bind({});
 Loading.args = {
     data: fetchingData,
-    options: {
-        ...options,
-    },
+    options,
 };
--- a/packages/visualizations/src/components/Table/Body.svelte
+++ b/packages/visualizations/src/components/Table/Body.svelte
@@ -2,18 +2,30 @@
     import type { Column, TableData } from './types';
     import Cell from './Cell';
 
+    export let isLoading: boolean | undefined;
+    export let loadingRowsNumber: number | undefined;
     export let columns: Column[];
     export let records: TableData;
 </script>
 
 <tbody>
-    {#each records as record}
-        <tr>
-            {#each columns as column}
-                <Cell rawValue={record[column.key]} {column} />
-            {/each}
-        </tr>
-    {/each}
+    {#if isLoading}
+        {#each Array(loadingRowsNumber) as _, index (index)}
+            <tr>
+                {#each columns as column}
+                    <Cell {isLoading} {column} />
+                {/each}
+            </tr>
+        {/each}
+    {:else}
+        {#each records as record}
+            <tr>
+                {#each columns as column}
+                    <Cell rawValue={record[column.key]} {column} />
+                {/each}
+            </tr>
+        {/each}
+    {/if}
 </tbody>
 
 <style>
--- a/packages/visualizations/src/components/Table/Body.svelte
+++ b/packages/visualizations/src/components/Table/Body.svelte
@@ -1,19 +1,18 @@
 <script lang="ts">
     import type { Column, TableData } from './types';
-    import Cell from './Cell';
+    import Cell, { LoadingCell } from './Cell';
 
-    export let isLoading: boolean | undefined;
-    export let loadingRowsNumber: number | undefined;
+    export let loadingRowsNumber: number | null;
     export let columns: Column[];
     export let records: TableData;
 </script>
 
 <tbody>
-    {#if isLoading}
-        {#each Array(loadingRowsNumber) as _, index (index)}
+    {#if loadingRowsNumber}
+        {#each Array(loadingRowsNumber) as _}
             <tr>
-                {#each columns as column}
-                    <Cell {isLoading} {column} />
+                {#each columns as __}
+                    <LoadingCell />
                 {/each}
             </tr>
         {/each}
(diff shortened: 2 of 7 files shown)


packages/visualizations-react/stories/Table/Table.stories.tsx
```diff
@@ -76,4 +81,10 @@ export const Unstyled = Template.bind({});
 Unstyled.args = {
     data,
     options: { ...options, unstyled: true },
+};
+
+export const Loading = Template.bind({});
+Loading.args = {
+    data: fetchingData,
+    options,
```
KevinFabre-ods, line 89: A remark: how many 'loading rows' should we display? On which prop we can use to get this number ? 

Is very common to have an empty `data` array when `loading` is true

cybervinvin: Skeleton must be displayed in each cell of the table. `data` can be an empty array, it is `isLoading` which conditions the display of skeletons.

etienneburdet (changes requested):
I've made a few code comments. I think we also have to update the `Paginated.stories.tsx` by modifying the `fetchData` utils and make it return `loading: false` during the timeout it sets, in order to better test the feature.

Else I just think the animation is waaaaay too fast and doesn't scale well in width. I think an homogene change of color and slower. To be seen with Meg.

I'm also in favor of adding a few more rows—say 10— just so that we are as close to a "normal" table as possible. It will change size a bit, but text around etc. should be a bit more in place. Another method would be to just a have one big SVG representing cells, it's more stable but doesn't instantiate tr/td.

packages/visualizations/src/components/Table/Cell/LoadingCell.svelte
```diff
@@ -0,0 +1,38 @@
+<td>
+    <div class="skeleton"></div>
+</td>
+
+<style>
+    :global(.ods-dataviz--default td) {
+        text-align: left;
+        padding: var(--spacing-75);
+    }
+    .skeleton {
```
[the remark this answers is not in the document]
cybervinvin: done

packages/visualizations/src/components/Table/Body.svelte
```diff
@@ -1,19 +1,29 @@
...
-    {/each}
+    {:else}
+        {#each records as record}
+            <tr>
+                {#each columns as column}
+                    <Cell rawValue={record[column.key]} {column} />
+                {/each}
+            </tr>
+        {/each}
+    {/if}
```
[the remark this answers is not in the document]
cybervinvin: done

packages/visualizations-react/stories/Table/Table.stories.tsx
cybervinvin: Now we add as many lines as the `recordsPerPage` if it exists, otherwise we add 5 lines.

cybervinvin:
**Warning**: the branch has been rebased.

the following changes have been added:
- replace the `LoadingCell` component by adding the animation into the `Cell` component (add new `isLoading` property)
- add as many "loading" lines as the value of `recordsPerPage` (if it does not exist, add 5 lines)
- improve the animation to reduce the speed

KevinFabre-ods (changes requested):
Looking good! 👏

Some code can be removed.

My main remark is about where the loading UI is handled. In my opinion, it should not be in the Cell component. Let's keep it with one responsibility: rendering formatted values.

packages/visualizations/src/components/Table/Body.svelte
```diff
@@ -2,18 +2,30 @@
...
-            {#each columns as column}
-                <Cell rawValue={record[column.key]} {column} />
-            {/each}
-        </tr>
-    {/each}
+    {#if isLoading}
+        {#each Array(loadingRowsNumber) as _, index (index)}
+            <tr>
+                {#each columns as column}
+                    <Cell {isLoading} {column} />
```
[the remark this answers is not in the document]
cybervinvin: I do not really understand. I understood that Etienne wanted to have only one component to display the data or the loading indicator ([here](https://github.com/opendatasoft/ods-dataviz-sdk/pull/240#discussion_r1621948967))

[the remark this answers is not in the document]
cybervinvin: We must display an animation for each column of the table, we always have the list of fields that build the columns of the table. I don't understand why this property `column` should be optional 🤔

packages/visualizations/src/components/Table/Cell/LoadingCell.svelte
```diff
@@ -0,0 +1,34 @@
...
+        content: '';
+        position: absolute;
+        top: 0;
+        left: 0;
+        width: 60%;
+        height: 100%;
+        background-color: #cbd2db;
+        filter: blur(26px);
+        animation: move 800ms linear 0s infinite alternate both;
+    }
```
etienneburdet, line 24: Can be `div { ... }` (but that's not really critical).

packages/visualizations/src/components/Table/Body.svelte
```diff
@@ -1,19 +1,30 @@
 <script lang="ts">
     import type { Column, TableData } from './types';
-    import Cell from './Cell';
+    import Cell, { LoadingCell } from './Cell';
 
+    export let loadingRowsNumber: number | null;
     export let columns: Column[];
```
etienneburdet, line 6:
```suggestion
    export let isLoading: boolean | undefined;
```

packages/visualizations/src/components/Table/Body.svelte
```diff
@@ -1,19 +1,30 @@
...
-    {/each}
+    {#if loadingRowsNumber}
+        {#each Array(loadingRowsNumber) as _}
+            <tr>
+                {#each columns as __}
+                    <LoadingCell />
+                {/each}
+            </tr>
+        {/each}
+    {:else}
```
etienneburdet, line 19: Doesn't take props.
```suggestion
                <LoadingCell />
```

etienneburdet (changes requested):
Couple TS comments (but TS already complains about thoses anyways), else looking good!

packages/visualizations/src/components/Table/Body.svelte
cybervinvin: 🤔

KevinFabre-ods:
LGTM ! 🚀 

@etienneburdet Could you re-review this PR?

KevinFabre-ods (approved):

etienneburdet (approved):

PR #241, opened 2024-05-29
fix: align sort number right and color sort field

etienneburdet:
## Summary

The goal for this PR is to align number header right, wether they are sortable or not (design request) and to inherit color for the button—else the sortable field would not inherit theme color.

### DIFF
--- a/packages/visualizations-react/stories/Table/custom-style.css
+++ b/packages/visualizations-react/stories/Table/custom-style.css
@@ -32,6 +32,10 @@
     color: #808080;
 }
 
+.design-system {
+    color: blue;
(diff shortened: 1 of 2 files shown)


KevinFabre-ods (approved):
LGTM

To test the inherit color, we could make a column sortable so that the custom color story apply this change

etienneburdet (commented):
I added some top level colors in the color sortable stories ✨

KevinFabre-ods (approved):
LGTM

I just rejected an baseline change because it was the choropleth tooltip and we know this snapshot is unstable

PR #242, opened 2024-05-31
Export svelte uncompiled components

etienneburdet:
## Summary

The goal for this PR is to export Svelte uncompiled components, so that they can be used **in a SvelteKit app**

### Changes
I added a `svelte` field in package.json

`exports: { ".": { "import": …; "svelte": …; }` seems to confuse rollup and vite (contrary to what the doc suggests). It's be explored in further tests though, this is the right way to export both ESM and umd, something that has been troublesome.

The explicit import of `registerable` is to avoid tree shaking. I deleted the export the namespace since we actually don't use it, I don't think it had been tested that it worked and was messing up with explicit imports.

### Changelog
It's now possible to import svelte uncompiled components to be used in a Svelte app: 
```
<script>
import { Chart } from '@opendatasoft/visualizations';
</script>

<Chart {data} {options} />
```

## Open discussion & test method
Soooooo… hardest one liner of my life. So trick is to have ChartJs work with both SSR and Treeshaking

To test if it's working, create a SvelteKit app and install visiualizations:
```
npm create svelte@latest my-app
cd my-app
npm install
npm install --save @opendatasoft/visualizations@0.24.1-beta.1
npm run dev -- --open
```

And try to create char. @fpassaniti you can find `data` and `options` to try with [here](https://616594e84c1692003af430f3-jombspzrzd.chromatic.com/?path=/docs/chart-axisassemblages--docs)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28540,7 +28540,7 @@
     },
     "packages/visualizations": {
       "name": "@opendatasoft/visualizations",
-      "version": "0.24.1",
+      "version": "0.24.1-beta.0",
       "license": "MIT",
       "dependencies": {
         "@placemarkio/geo-viewport": "^1.0.2",
@@ -28613,10 +28613,10 @@
     },
     "packages/visualizations-react": {
       "name": "@opendatasoft/visualizations-react",
-      "version": "0.24.1",
+      "version": "0.24.1-beta.0",
       "license": "MIT",
       "dependencies": {
-        "@opendatasoft/visualizations": "0.24.1",
+        "@opendatasoft/visualizations": "0.24.1-beta.0",
         "use-callback-ref": "^1.2.4"
       },
       "devDependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -3048,9 +3048,8 @@
     },
     "node_modules/@lerna/create/node_modules/@nrwl/tao": {
       "version": "18.3.5",
-      "resolved": "https://registry.npmjs.org/@nrwl/tao/-/tao-18.3.5.tgz",
-      "integrity": "sha512-gB7Vxa6FReZZEGva03Eh+84W8BSZOjsNyXboglOINu6d8iZZ0eotSXGziKgjpkj3feZ1ofKZMs0PRObVAOROVw==",
       "dev": true,
+      "license": "MIT",
       "dependencies": {
         "nx": "18.3.5",
         "tslib": "^2.3.0"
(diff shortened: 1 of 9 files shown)


etienneburdet:
> 👋 @etienneburdet I'm struggle to render a viz in a Svelte kit project.
> 
> First, I had a `self is not defined` error for the page where I import the visualizations module. To resolve it I set `export const ssr = false;` in the javascript file associated to the page. (This error seems fair, can't render a graph on the server side)
> 
> Then, I had missing dependency error: `saas is not defined`. I had to install `saas` as dev dependency
> 
> Finally, I'm stuck with a wrong export error: `import { geoPath, GeoProjection } from "d3-geo";` `No matching export in "node_modules/d3-geo/src/index.js" for import "GeoProjection"`
> 
> Did you experience any of it?

I had these at some point indeed… Seems like I didn't commit everything then. I'll have a look!

etienneburdet:
Should be good for testing on 0.24.1-beta.2 @fpassaniti @KevinFabre-ods  
I moved sass as a dep (since it's needed on the end user now that we import components with sass) and seperate type import since it wasns't recognized as a TS thing.

KevinFabre-ods:
> 0.24.1-beta.2

OK so new error: 
> TypeError: parent_component is undefined

Did you successfully render a chart? What is your code? 
What I'm testing: 

```js
<script>
    import {Chart} from "@opendatasoft/visualizations";

    // ...
</script>

<Chart {...args} />
```

manondravet:
Hello, I rendered a chart on my side, just re-using the code Étienne mentioned:

`<script>
import { Chart } from '@opendatasoft/visualizations';
</script>

<Chart {data} {options} />`

worked with both a bar chart stacked and an area chart stacked

etienneburdet:
> OK so new error:
> 
> > TypeError: parent_component is undefined
> 
> Did you successfully render a chart? What is your code? What I'm testing:
> 
> ```js
> <script>
>     import {Chart} from "@opendatasoft/visualizations";
> 
>     // ...
> </script>
> 
> <Chart {...args} />
> ```

If you had beta1 before, you have to delete your node modules and reinstall. It will remove vite prebundled deps. That may be it… maybe not 😬

KevinFabre-ods:
Thanks I'll try that

KevinFabre-ods (approved):
I could render every visualizations 🎊 

Before merging this PR, I think we can clean up edits in `package.json` files to remove the change of version. 

0.24.1 is already out. This PR will be in the 0.25.0 version

KevinFabre-ods:
@fpassaniti and @manondravet This feature is available in the latest version ➡️ 0.25.0

PR #243, opened 2024-05-31
[FIX][TABLE][SC-50479] Update page size

KevinFabre-ods:
## Summary

The goal for this PR is to fix the behavior for the page size select as it can't be used on Chrome/Safari

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-50479](https://app.shortcut.com/opendatasoft/story/50479).

### Changes
Use the `input` event instead of the `select`. I didn't dive into the why, but it does the job.

### DIFF
--- a/packages/visualizations/src/components/Pagination/PageSize.svelte
+++ b/packages/visualizations/src/components/Pagination/PageSize.svelte
@@ -17,7 +17,7 @@
     }
 </script>
 
-<select bind:value on:change={(e) => onChange(toNumber(e.currentTarget.value))}>
+<select bind:value on:input={(e) => onChange(toNumber(e.currentTarget.value))}>
     {#each options as option}
         <option label={option.label} value={option.value} />
     {/each}


etienneburdet:
Don't why indeed, but it works and all three browsers. We could add an `-webkit-appearance: none;` too in order to have the same styling in Safari (right now it has the gloss & stuff)

KevinFabre-ods:
> Don't know why indeed, but it works and all three browsers. We could add an `-webkit-appearance: none;` too in order to have the same styling in Safari (right now it has the gloss & stuff)

Using webkit-appearance: none will remove the arrow in all browsers. 
There is a thread on Stackoverflow to mitigate that. (https://stackoverflow.com/questions/7638677/how-can-i-remove-the-gloss-on-a-select-element-in-safari-on-mac)

I'm torn between keeping like this or adding extra CSS that will make CSS override trickier.

Best solution if to use a custom component for select inputs. But do we really want to do that now :/

KevinFabre-ods:
As tomorrow we want to release the table for some clients, we merge as it is. We'll fix the style in a other story right after

PR #244, opened 2024-06-11
fix: remove safari select style

etienneburdet:
## Summary

The goal for this PR is to remove safari glossy style on the page size select

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-50489](https://app.shortcut.com/opendatasoft/story/50489).

### Changes
The solution is `-webkit-appearance: none` and readd the arrows as background image. The problem is, the background image is for all browsers, so we need to have `appearance: none;` globally.

This only applies to default style.

### Changelog
PageSize select doesn't have the default safari style anymore.

## Review checklist

- [x] Description is complete

### DIFF
--- a/packages/visualizations/src/components/Pagination/PageSize.svelte
+++ b/packages/visualizations/src/components/Pagination/PageSize.svelte
@@ -23,6 +23,8 @@
     {/each}
 </select>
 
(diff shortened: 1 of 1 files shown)


KevinFabre-ods (approved):
LGTM 🟢

PR #245, opened 2024-06-11
[FEAT][EXPLO][SC-46279] Table - Geo field

KevinFabre-ods:
## Summary

The goal for this PR is to explore how we can implement the table geo field

https://github.com/opendatasoft/ods-dataviz-sdk/assets/117300300/95c30706-13df-4949-ab55-83f472c747fb

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-46279](https://app.shortcut.com/opendatasoft/story/46279).

### Changes

Did a lot of refactoring because my main idea was to re-use some part of the code that composes `MapPoi`. So I extracted a Svelte component (placed it in `Map/WebGl`) that can be used as a common template to generate maps. It is currently used by the `PoiMap` and `Table` blocks. 

There should no visual and functional changes for the PoiMap visualization.

So, the table has a new column: GeoColumn with a lot of options:

- `display`: a function based on the cell value to render a text from complexe object (coordinates, shapes, etc...)
- `mapOptions`: an object where we configure render parameters for the map (like its style, center, bbox, interactiveness, etc...). In our application, this is where we pass the basemap style.
- `layers`: a function to render [map layers](https://maplibre.org/maplibre-style-spec/layers/) based on the cell value. 
- `sources`: a function to render [map sources](https://maplibre.org/maplibre-style-spec/sources/) based on the call value

Currently, the `WelGL` map component only supports `circle` and `symbol` layers. If we go in this direction, we need to add support for the fill layers. To render geoshapes in our application. Done in https://github.com/opendatasoft/ods-dataviz-sdk/pull/246

In this PR, we also add aliases for visualizations paths

- [x] Description is complete


etienneburdet:
Didn't fully review yet, but just a quick factor comment. Firts, the ChoroplethMap rename was long due… thanks! I don't think we need to extract the `Map.svelte` into a `utils` folder—they usually end up being a mixed bag of everything. As it is it's fine in PoiMap, GeoCell can totally import from there and it's not applicable as is for other maps.

The main factor we need now is to move to `transformStyle` which should simplify a lot of things, including a common `MapRender.svelte`, which could lead to something like this: 
```
Map
├── WegGl.svelte // can be imported in table
├── Svg.svelte
└── Choropleth
    ├── ChoroplethSvg.svelte // imports Svg
    ├── ChoroplethVectorTiles.svelte // imports WebGl
    ├── ChoroplethGeojson.svelte // imports WebGl
└── Poi
    ├── Poi.svelte // can be imported in Table // imports WebGl
```
That was pretty much the thinking at the time (hence the /Map/Choropleth folder) and I think it has quite a few advantages, especially in how we avoid doubling maplibre implementations.

KevinFabre-ods:
> Didn't fully review yet, but just a quick factor comment. Firts, the ChoroplethMap rename was long due… thanks! I don't think we need to extract the `Map.svelte` into a `utils` folder—they usually end up being a mixed bag of everything. As it is it's fine in PoiMap, GeoCell can totally import from there and it's not applicable as is for other maps.
> 
> The main factor we need now is to move to `transformStyle` which should simplify a lot of things, including a common `MapRender.svelte`, which could lead to something like this: 
> 
> ```
> Map
> ├── WegGl.svelte // can be imported in table
> ├── Svg.svelte
> └── Choropleth
>     ├── ChoroplethSvg.svelte // imports Svg
>     ├── ChoroplethVectorTiles.svelte // imports WebGl
>     ├── ChoroplethGeojson.svelte // imports WebGl
> └── Poi
>     ├── Poi.svelte // can be imported in Table // imports WebGl
> ```
> 
> That was pretty much the thinking at the time (hence the /Map/Choropleth folder) and I think it has quite a few advantages, especially in how we avoid doubling maplibre implementations.


 
@etienneburdet  I see where you want to go. Looks good. This is not light work to do. We might want to start with the current file organization to fulfill our table story scope by the end of the month.

After that, we can start with the refactoring.

Currently, we have a `Map.ts` file that is responsible for instantiating a Maplibre instance and defining POI behaviors (hover on feature, click on feature) and other more common behaviors like popup displays, reactive properties (center, bbox, zoom, layers, sources, interactivity).

How do you see the refactor for this file?

packages/visualizations/src/components/Table/Cell/Format/GeoFormat.svelte
```diff
@@ -0,0 +1,42 @@
+<script lang="ts">
+    import Map from '../../../utils/Map';
+    import type { MapOptions, MapData } from '../../../utils/Map/types';
```
etienneburdet, line 3: I think we can setup `src` or even `utils` aliases same as we do in `visualizations-react`

etienneburdet (commented):
So beside the fact I would put the components in `Map` rather than `utils`—which isn't critical either—, the grouping itself loosk good and works well 👍

The geoformat field works pretty well. UX-wise, I think I would be more for a modal/popin that centers in the middle of the table (or where we click). This way we don't risk breaking the other layouts. Could be for a next ticket though. I also have in mind the same kind of pattern for images: thumbnail that we can enlarge as a popup. Svelte transitions could definitely be of use here.

As for the `Map.ts` refactor, first and foremost, since we mainly don't touch the old code, but just break it appart, I think we can keep that for another PR. But my idea would be something like:
```
// Maplibre.svelte
<script>
setContext(map);
</script>

{#if map && maready}
  <slot />
{/if}

// Choropleth.svelte
<script>
export let options
const map = getContext(map);

onMount(() => initializeChoropleth(options);
</script>

<PopUp />
<Controls />

//ChoroplethVectorTiles
<Maplibre>
  <Choropleth {vectorTilesOptions} />
<Maplibre
```

The idea here would be that everything that requires `enqueue` is in the slot, so that it can be transformed into pure JS functions, that are called in `initilialize`. Everything that create HTML, is its own component (again guarded with the slot the guarentees loading). And then all reactive stuff (mainly layers) should belong to `transformStyle` if I'm correct.

packages/visualizations/src/components/PoiMap/MapPoi.svelte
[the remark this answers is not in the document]
RafaelSzmarowski: Or just `Poi.svelte` but the folder should be `PoiMap` if we look at `ChoroplethMap`

RafaelSzmarowski (commented):
Nice work ! The GeoFormat cell is easy to read and understand even when discovering the code for the first time !

I agree with @etienneburdet to avoid the utils folder but it’s not critical and can be made later indeed. By the way, nice suggestion to use the slots, haven’t tested it but if it works it’s nice and clean. 

I haven’t followed the UX discussions around the map display inside the table so I don’t know the final features of the table cell and I might be wrong: 

IMO if the cell will be small and static like now, I automatically think about something else than a webGL (have you thought about a png or svg generated through map.getCanvas().toDataURL ? I don't know the impact in terms of performance...) but if the cell must be interactive then it should be a Maplibre webGL and probably in an expandable modal / popup as the current size is too small to have interactivity.

The webGL browser limit could be reached very quickly : 

https://github.com/opendatasoft/ods-dataviz-sdk/assets/84501475/9aab24b1-e8d2-40aa-b38a-21d6b43982e0


I guess it could be dealt with by displaying only one map at a time in a bigger modal / popup or if we keep the current implementation, maybe by a better handling of `webglcontextlost` to avoid having broken cells like in my screen recording.

A basic listener could do the job but there might be a better solution: 
in `GeoFormat.svelte`
```
    import { onMount, afterUpdate } from 'svelte';
    import Map from '../../../utils/Map';
    import type { MapOptions, MapData } from '../../../utils/Map/types';

    export let rawValue: unknown;
    export let display = (v: unknown) => v;
    export let mapOptions: MapOptions;
    export let sources: (v: unknown) => MapData['sources'] = () => ({});
    export let layers: (v: unknown) => MapData['layers'] = () => [];

    let showMap = false;
    let mapContainer: HTMLDivElement | null = null;

    $: data = {
        sources: sources(rawValue),
        layers: layers(rawValue),
    };

    function onLabelClick() {
        showMap = !showMap;
    }

    function handleWebGLContextLost(event: Event) {
        // Prevent the default behavior of the event
        event.preventDefault();
        // Close the map component properly
        showMap = false;
    }

    function setupWebGLContextListener() {
        if (mapContainer) {
            const canvas = mapContainer.querySelector('canvas');
            if (canvas) {
                canvas.addEventListener('webglcontextlost', handleWebGLContextLost, false);
            }
        }
    }

    onMount(() => {
        if (showMap) {
            setupWebGLContextListener();
        }
    });

    afterUpdate(() => {
        if (showMap) {
            setupWebGLContextListener();
        }
    });
</script>

<div>
    <!-- svelte-ignore a11y-click-events-have-key-events -->
    <div class="label" on:click={onLabelClick}>{display(rawValue)}</div>
    {#if showMap}
        <div class="table-cell-map-container" bind:this={mapContainer}>
            <Map options={mapOptions} {data} />
        </div>
    {/if}
</div>

<style>
    .label {
        cursor: pointer;
        text-decoration: underline;
    }
    .table-cell-map-container {
        width: 100%;
        height: 250px;
    }
</style>

```

KevinFabre-ods:
Thank you @etienneburdet and @RafaelSzmarowski  for your reviews 


I'll update and move all files related to maps into the /Map folder.

As for the UX, I'm waiting on the design. I don't want to dive into an implementation that doesn't align with what the design team is planning.

Regarding the WebGL context loss, I'm also waiting for the design. Perhaps they will implement a feature where only one map can be displayed at a time.

About using other resources than Webgl to display a map, I considered it but didn't pursue it because:
-  Attributions are not visible when an image is generated from map.getCanvas.
- We need to ensure that the layers have been loaded and are visible before calling map.getCanvas.
- IMO, if we want to use images instead of WebGL maps, the images should be generated server-side to avoid layers loading issues as described. With the image column type, you can display these images.

etienneburdet:
I think `map.getCanvas().toDataURL` will have to pop a webgl instance anyways (before rexporting it to image), so it might not solve the problem if we don't save images somewhere. I'm very much in favor of "only one mini map at a time", both for UX and performance.

packages/visualizations/src/components/Table/Cell/Format/GeoFormat.svelte
KevinFabre-ods: Done https://github.com/opendatasoft/ods-dataviz-sdk/pull/245/commits/f24f939ad6a953924473c5e72d9182cde9428593

KevinFabre-ods:
@etienneburdet & @RafaelSzmarowski 

This PR is ready for a second round of review. 
Since your last review: 
- aliases have been added for visualizations package folders
- A minimap can be previewed (in a tooltip)when user focus or hover a geo cell

packages/visualizations/src/components/Table/Cell/Format/GeoFormat.svelte
```diff
@@ -0,0 +1,62 @@
...
+        theme: 'ods-visualization-table',
+        delay: [500, 0],
+        duration: [275, 0],
+        maxWidth: 'none',
+        onTrigger: (instance) => {
+            showMap = true;
+        },
+        onUntrigger: (instance) => {
+            showMap = false;
+        },
```
KevinFabre-ods, line 37: Only way that I found to render properly the tooltip.
Without it, the tooltip is not visible when events (focus, mousenter) are triggered

packages/visualizations/src/components/Table/Cell/Format/GeoFormat.svelte
```diff
@@ -0,0 +1,62 @@
+<script lang="ts">
+    import { tick } from 'svelte';
```
RafaelSzmarowski, line 2: I don't see where this is used ? Probably a forgotten import.

packages/visualizations/src/components/Table/Cell/Format/GeoFormat.svelte
```diff
@@ -0,0 +1,62 @@
...
+<div
+    role="button"
+    tabindex="0"
+    use:tippy={{
+        content: tooltipContent,
+        theme: 'ods-visualization-table',
+        delay: [500, 0],
+        duration: [275, 0],
+        maxWidth: 'none',
+        onTrigger: (instance) => {
```
RafaelSzmarowski, line 32: Do you need instance as an arg here ?

packages/visualizations/src/components/Table/Cell/Format/GeoFormat.svelte
```diff
@@ -0,0 +1,62 @@
...
+    import type { WebGlMapOptions, WebGlMapData } from 'components/Map';
+
+    export let rawValue: unknown;
+    export let display = (v: unknown) => v;
+    export let mapOptions: WebGlMapOptions;
+    export let sources: (v: unknown) => WebGlMapData['sources'] = () => ({});
+    export let layers: (v: unknown) => WebGlMapData['layers'] = () => [];
+
+    let tooltipContent;
+    let showMap;
```
RafaelSzmarowski, line 15: It would probably be better to define the types here, I don't know if I have successfully updated my TS config but I have TS errors related.

packages/visualizations/src/components/Table/Cell/Format/GeoFormat.svelte
```diff
@@ -0,0 +1,62 @@
...
+    use:tippy={{
+        content: tooltipContent,
+        theme: 'ods-visualization-table',
+        delay: [500, 0],
+        duration: [275, 0],
+        maxWidth: 'none',
+        onTrigger: (instance) => {
+            showMap = true;
+        },
+        onUntrigger: (instance) => {
```
RafaelSzmarowski, line 35: Same as previously, seems unused.

packages/visualizations/src/components/Table/Cell/Format/GeoFormat.svelte
```diff
@@ -0,0 +1,62 @@
...
+</div>
+
+<style lang="scss">
+    .label {
+        cursor: pointer;
+        text-decoration: underline;
+    }
+    :global(.table-cell-map-container) {
+        width: 360px;
+        height: 240px;
```
RafaelSzmarowski, line 57: A fixed value seems not like a long term solution. How we gonna deal with responsiveness here ? Is it left for a future PR ?

RafaelSzmarowski (approved):
Nice work ! As far as I have tested it, it looks clean and nice, way better UX in my opinion than displaying the image inside the cell. 

I have left only non blocking minor comments.

etienneburdet (commented):
Working well on my side! I just need to look into this TS error before I validate. 

I assume the idea is to move Choroplethmap back into the new Map folder and using the WebGL.svelte for both. I'm fine with, we'll just have to be careful with TS alsiases when importing in the meantime.

And I also assume we make another story/PR for the modal (which seems a good idea indeed).

packages/visualizations/src/components/Map/WebGl/constants.ts
```diff
@@ -25,24 +23,35 @@ export const DEFAULT_SORT_KEY_VALUE: DataDrivenPropertyValueSpecification<number
...
 export const POPUP_WIDTH = 300;
 
 export const CONTROL_POSITION: ControlPosition = 'top-right';
 
 // Update styles in ./MapRender.svelte if one of these classnames must change.
-const POPUP_CLASSNAME = 'poi-map__popup';
-export const POPUP_FEATURE_CONTENT = 'poi-map__popup-feature-content';
-export const POPUP_FEATURE_CONTENT_LOADING = 'poi-map__popup-feature-content--loading';
-export const POPUP_NAVIGATION_CONTROLS_CLASSNAME = 'poi-map__popup-navigation-controls';
+const POPUP_CLASSNAME = 'ods-visualization__map-popup';
```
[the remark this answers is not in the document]
etienneburdet: Since we're beta, the API change seems fair enough. Not sure it's fully needed though, we have an `ods-visualization-poi-map` that wraps everything (but maybe the popup is out of this tree?).

packages/visualizations/src/index.ts
```diff
@@ -1,21 +1,22 @@
...
-export * from './components/Legend/types';
-export * from './components/Table/types';
-export * from './components/Pagination/types';
+export * from 'components/Map';
+
+export * from 'types';
+export * from 'components/types';
+export * from 'components/Chart/types';
+export * from 'components/KpiCard/types';
+export * from 'components/ChoroplethMap/types';
```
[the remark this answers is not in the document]
RafaelSzmarowski: PoiMap types seems to be exported directly from components/Map if I'm not misunderstanding ?

packages/visualizations/src/components/Table/Cell/Format/GeoFormat.svelte
KevinFabre-ods: Yes, for the moment we go with that.
We didn't get any design about responsiveness

packages/visualizations/src/index.ts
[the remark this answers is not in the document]
KevinFabre-ods: Yes, types are exported from `components/Map`

packages/visualizations/src/components/Map/WebGl/constants.ts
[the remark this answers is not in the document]
KevinFabre-ods: Adding classes to the popup is required to handle our 3 variants (tooltip, sidebar and modal)

However, now the map canvas is wrapper around a container (`.ods-visualization__map-container`)
So we can use it to scope our style. 

Instead of doing 
`:global(.ods-visualization__map-popup.ods-visualization__map-popup--as-sidebar)`
we can do 
`.ods-visualization__map-container :global(.maplibregl-popup--as-sidebar),`

See https://github.com/opendatasoft/ods-dataviz-sdk/pull/245/commits/5b42d5d118b92486cf86c288d4eb45e9efd6120b

It limit a bit the usage of custom classnames

packages/visualizations/src/components/Table/Cell/Format/GeoFormat.svelte
KevinFabre-ods: Yes fixed here
https://github.com/opendatasoft/ods-dataviz-sdk/pull/245/commits/e7e8a4ed09953ea970f68dcb6db8144c45897773

packages/visualizations/src/components/Table/Cell/Format/GeoFormat.svelte
KevinFabre-ods: nop, fixed here 
https://github.com/opendatasoft/ods-dataviz-sdk/pull/245/commits/e7e8a4ed09953ea970f68dcb6db8144c45897773

packages/visualizations/src/components/Table/Cell/Format/GeoFormat.svelte
KevinFabre-ods: Fixed 
https://github.com/opendatasoft/ods-dataviz-sdk/pull/245/commits/e7e8a4ed09953ea970f68dcb6db8144c45897773

packages/visualizations/src/components/Table/Cell/Format/GeoFormat.svelte
KevinFabre-ods: Fixed
https://github.com/opendatasoft/ods-dataviz-sdk/pull/245/commits/e7e8a4ed09953ea970f68dcb6db8144c45897773

packages/visualizations/src/components/Map/WebGl/WebGl.svelte
```diff
@@ -0,0 +1,261 @@
+<svelte:options immutable={true} />
```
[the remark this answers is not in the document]
KevinFabre-ods: Remove here https://github.com/opendatasoft/ods-dataviz-sdk/pull/245/commits/bb45dbd4c784ad9ed5e851093e16d2799f5563be

KevinFabre-ods:
Thanks for your review.

Indeed, there were TypeScript errors in the `GeoFormat.svelte `component. Sorry about that. They should be fixed now. However, the CI should have caught these errors.

In my opinion, we were missing an ESLint check for the Svelte components. This has been fixed in [this commit](https://github.com/opendatasoft/ods-dataviz-sdk/pull/245/commits/26bf616cefaf39cc3a5a047cb225deb737c91e67). I checked that these TypeScript errors were caught by the new test.

As we discussed this morning, we have an issue when building the Svelte package. The introduction of aliases breaks the paths for the TypeScript declaration files. @RafaelSzmarowski, this morning, with @etienneburdet, we tested various Rollup plugins to fix it. We were unsuccessful. So, we have decided to try upgrading Rollup to the latest version if it is quick to do.

If not, we will revert the alias commit [f24f939ad6a953924473c5e72d9182cde9428593](https://github.com/opendatasoft/ods-dataviz-sdk/pull/245/commits/f24f939ad6a953924473c5e72d9182cde9428593) to avoid blocking this PR and to address the Rollup update in a dedicated PR.

We will sum up tomorrow morning to check the progress of the upgrade and decide what to do in this PR.

etienneburdet (approved):
All good for me! I like the new classes better indeed too.

Svelte check should do, but I'm surprised it didn't work, we used to have implicit any errors on uninitialized variables. But it works… 🤷‍♂️

cybervinvin (approved):
I arrive after the battle but, in any case, I don't have much to say on a subject that I don't really understand.
Now, overall the code seems easier to read and that's generally a good sign.
However, I have 2 questions on very small points of detail which are absolutely not blocking.

Bravo and thanks for the work 👏🏼

packages/visualizations/src/components/Map/Poi/types.ts
```diff
@@ -0,0 +1,20 @@
+import type { Async } from '../../../types';
```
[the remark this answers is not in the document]
KevinFabre-ods: 🤔 Why ? 
It is used for at line 6 

https://github.com/opendatasoft/ods-dataviz-sdk/blob/9ae11aaf5724b56a09206309de68a0e0ea097ee7/packages/visualizations/src/components/Map/Poi/types.ts#L6

packages/visualizations/src/components/Map/WebGl/WebGl.svelte
```diff
@@ -0,0 +1,257 @@
...
+    .ods-visualization__map-container :global(.maplibregl-popup--as-modal) {
+        width: 100% !important;
+    }
+    .ods-visualization__map-container :global(.maplibregl-popup--as-sidebar) {
+        /* 300px for content and 26px for padding */
+        width: calc(300px + 26px) !important;
+    }
+
+    .ods-visualization__map-container :global(.maplibregl-popup--as-tooltip) {
+        width: 300px !important;
```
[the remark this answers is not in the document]
KevinFabre-ods: I was able to remove most of the !important style expect for the transform as it is set directly on the Node right after the popup opens

https://github.com/maplibre/maplibre-gl-js/blob/b63b28454a7c43341d41211c0e9c63948597e8ee/src/ui/popup.ts#L647

PR #247, opened 2024-06-14
feat: add an empty state for the table

etienneburdet:
## Summary

The goal for this PR is to add a small phrase to indicate that no records were found—but appart from that the table hasn't crashed.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-50771](https://app.shortcut.com/opendatasoft/story/50771).

### Changes
I added another alternative render for the body, similar to the loading state.

I could have made it into it's own component, but I don't find the `Body.svelte` too cluttered for now. I'm totally open to it though.

### DIFF
--- a/packages/visualizations-react/stories/Table/Table.stories.tsx
+++ b/packages/visualizations-react/stories/Table/Table.stories.tsx
@@ -77,3 +77,9 @@ Unstyled.args = {
     data,
     options: { ...options, unstyled: true },
 };
+
+export const emptyState = Template.bind({});
+emptyState.args = {
+    data: { value: [], loading: false },
+    options: { ...options, emptyStateLabel: 'Neniuj registroj trovitaj...' },
+};
--- a/packages/visualizations/src/components/Table/Body.svelte
+++ b/packages/visualizations/src/components/Table/Body.svelte
@@ -9,7 +9,7 @@
 </script>
 
 <tbody>
-    {#if records.length === 0}
+    {#if records.length === 0 && !loadingRowsNumber}
         <tr>
             <td colspan={columns.length}>
                 <em>{emptyStateLabel}</em>
--- a/packages/visualizations/src/components/Table/Body.svelte
+++ b/packages/visualizations/src/components/Table/Body.svelte
@@ -4,9 +4,17 @@
 
     export let columns: Column[];
     export let records: TableData;
+    export let emptyStateLabel = 'No records found...';
 </script>
 
 <tbody>
+    {#if records.length === 0}
+        <tr>
+            <td colspan={columns.length}>
+                <em>{emptyStateLabel}</em>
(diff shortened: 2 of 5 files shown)


packages/visualizations-react/stories/Table/Table.stories.tsx
```diff
@@ -76,4 +76,10 @@ export const Unstyled = Template.bind({});
 Unstyled.args = {
     data,
     options: { ...options, unstyled: true },
+};
+
+export const emptyState = Template.bind({});
+emptyState.args = {
+    data: { value: [], loading: false },
+    options: { ...options, emptyStateLabel: 'Neniuj registroj trovitaj...' },
```
KevinFabre-ods, line 84: Story should show 'Neniuj registroj trovitaj...'

Prop drilling to `Body.svelte` is missing

KevinFabre-ods (changes requested):
Look goods, we just need to extract it from the options in `TableCard.svelte`

KevinFabre-ods:
@etienneburdet With the latest merges, we have some conflicts.
Could you please resolve them, so that we can do a release ? 

Thanks

packages/visualizations-react/stories/Table/Table.stories.tsx
```diff
@@ -83,8 +83,8 @@ Unstyled.args = {
...
 };
 
-export const Loading = Template.bind({});
-Loading.args = {
-    data: fetchingData,
-    options,
+export const emptyState = Template.bind({});
+emptyState.args = {
+    data: { value: [], loading: false },
+    options: { ...options, emptyStateLabel: 'Neniuj registroj trovitaj...' },
```
KevinFabre-ods, line 89: We lost the "loading state" story

KevinFabre-ods (approved):
LGTM ✅  

We just lost the loading story during the merge

packages/visualizations-react/stories/Table/Table.stories.tsx
etienneburdet: should be good 🙈

packages/visualizations/src/components/Table/Body.svelte
```diff
@@ -5,9 +5,17 @@
...
 </script>
 
 <tbody>
+    {#if records.length === 0}
+        <tr>
+            <td colspan={columns.length}>
+                <em>{emptyStateLabel}</em>
+            </td>
+        </tr>
+    {/if}
```
KevinFabre-ods, line 18: We need to display this UI when `isLoading` is true
image

Thanks to the loading story 😝

etienneburdet (commented):
haaaa indeed, pushed a bit too fast.

KevinFabre-ods (approved):
LGTM

PR #249, opened 2024-06-21
Add direction on the `Table` component

cybervinvin:
## Summary

The goal for this PR is to change the reading direction in the `Table` component depending on a property.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-48044](https://app.shortcut.com/opendatasoft/story/48044/sdk-add-direction-rtl-to-table).

### Changes

Add a new property `direction` for `TableCard`.
Use 2 new stores: `direction` and `isRtl` to propagate the property in the component tree.

#### Breaking Changes

None

### Changelog

In the Table component: It's now possible to change the reading direction depending on the locale.

## To be tested

In storybook: change the `direction` value in `options`

## Review checklist

- [X] Description is complete
- [X] Code is ready for a release on NPM

### DIFF
--- a/packages/visualizations-react/stories/Table/options.ts
+++ b/packages/visualizations-react/stories/Table/options.ts
@@ -82,6 +82,7 @@ const options: TableOptions = {
         href: '',
     },
     locale: 'fr',
+    direction: 'ltr',
 };
 
 export default options;
--- a/packages/visualizations/src/components/Table/Cell/Cell.svelte
+++ b/packages/visualizations/src/components/Table/Cell/Cell.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
     import type { Column } from '../types';
     import Format, { isValidRawValue } from './Format';
+    import { isRtl } from '../store';
 
     export let rawValue: unknown;
     export let column: Column;
@@ -9,7 +10,7 @@
 </script>
 
 <!-- To display a format value, rawValue must be different from undefined or null -->
-<td class={`table-data--${dataFormat}`}>
+<td class="{`table-data--${dataFormat}`} {$isRtl ? 'rtl-direction' : ''}">
     {#if isValidRawValue(rawValue)}
         <svelte:component this={Format[dataFormat]} {rawValue} {...options} />
     {/if}
@@ -19,6 +20,10 @@
     :global(.ods-dataviz--default td) {
         text-align: left;
         padding: var(--spacing-75);
+    }
+
+    :global(.ods-dataviz--default td.rtl-direction) {
+        text-align: right;
     }
 
     :global(.ods-dataviz--default td.table-header--number) {
--- a/packages/visualizations/src/components/Table/Headers/Headers.svelte
+++ b/packages/visualizations/src/components/Table/Headers/Headers.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
     import SortButton from './SortButton.svelte';
+    import { isRtl } from '../store';
     import type { Column } from '../types';
 
     export let columns: Column[];
@@ -8,7 +9,7 @@
 <thead>
     <tr>
         {#each columns as column}
-            <th class={`table-header--${column.dataFormat}`}>
+            <th class="{`table-header--${column.dataFormat}`} {$isRtl ? 'rtl-direction' : ''}">
                 {#if column.onClick}
                     <SortButton
                         sorted={column?.sorted}
@@ -32,6 +33,10 @@
     :global(.ods-dataviz--default th) {
         text-align: left;
         padding: var(--spacing-75);
+    }
+
+    :global(.ods-dataviz--default th.rtl-direction) {
+        text-align: right;
     }
 
     :global(.ods-dataviz--default th.table-header--number) {
--- a/packages/visualizations/src/components/Table/Table.svelte
+++ b/packages/visualizations/src/components/Table/Table.svelte
@@ -4,6 +4,7 @@
     import { generateId } from '../utils';
     import Headers from './Headers';
     import Body from './Body.svelte';
+    import { isRtl } from './store';
 
     export let loadingRowsNumber: number | null;
     export let columns: Column[];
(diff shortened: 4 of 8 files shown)


packages/visualizations-react/stories/Table/options.ts
KevinFabre-ods, line 1: image

title, subtitle and pagination are not in RTL configuration.

To align title and subtitle to the right we need to pass a prop to the `Card` component. I really don't know if we want to do that.

KevinFabre-ods, line 1: You can add this storybook addon (https://storybook.js.org/addons/storybook-addon-rtl/) to test components based on the page direction

packages/visualizations/src/components/Table/Cell/Cell.svelte
```diff
@@ -21,6 +22,10 @@
         padding: var(--spacing-75);
     }
 
+    :global(.ods-dataviz--default td.rtl-direction) {
+        text-align: right;
+    }
```
KevinFabre-ods, line 27: Using `css` to set direction seems not recommended

https://www.w3.org/International/questions/qa-html-dir.en.html#no_css

packages/visualizations/src/components/Table/types.ts
```diff
@@ -101,6 +101,8 @@ export type TableOptions = {
     source?: Source;
     /** To format date and number with the right locale. Default is from browser language */
     locale?: string;
+    /** To set the reading direction of the table (based on locale). Default is 'ltr' */
+    direction?: string;
```
KevinFabre-ods, line 105:
```suggestion
    direction?: 'ltr' | 'rtl'; 
```
To be more precise

KevinFabre-ods (commented):
I have left some general comments.

As we discussed, we are awaiting confirmation on how the direction should be handled for this component:
 - Based on the HTML attribute `dir`
 - Or by using a `direction` prop

In my opinion, we should rely solely on the HTML attribute `dir` to avoid the complexity of managing potential difference between the `dir` attribute and the `direction` prop.

packages/visualizations/src/components/Table/Table.svelte
```diff
@@ -31,6 +35,10 @@
         display: none;
     }
 
+    .rtl-direction {
+        direction: rtl;
+    }
```
etienneburdet, line 40: I think `<table dir="rtl">` has more benefits (a11y, send to cells etc.)

packages/visualizations/src/components/Table/Cell/Cell.svelte
```diff
@@ -21,6 +22,10 @@
         padding: var(--spacing-75);
     }
 
+    :global(.ods-dataviz--default td.rtl-direction) {
```
etienneburdet, line 25: From my tries, if we have `<table dir="rtl">` then text in td will be rtl. We might want to force ltr for some formats though (number, maybe URL).

packages/visualizations/src/components/Table/Headers/Headers.svelte
```diff
@@ -34,6 +35,10 @@
         padding: var(--spacing-75);
     }
 
+    :global(.ods-dataviz--default th.rtl-direction) {
+        text-align: right;
+    }
```
etienneburdet, line 40:
```suggestion
    :global(.ods-dataviz--default table[dir="rtl"] th) {
        text-align: right;
    }
```

packages/visualizations/src/components/Table/TableCard.svelte
```diff
@@ -20,10 +21,13 @@
...
         unstyled,
         locale: localeOption,
+        direction: directionOption,
         pagination,
     } = options);
     $: $locale = localeOption || navigator.language;
     $: defaultLoadingRowsNumber = pagination ? pagination.recordsPerPage : 5;
+    $: $direction = directionOption || (RTL_LANGUAGES_CODE.includes($locale) ? 'rtl' : 'ltr');
+    $: $isRtl = $direction === 'rtl';
     $: loadingRowsNumber = isLoading ? defaultLoadingRowsNumber : null;
```
etienneburdet, line 31: You could do with only one store here, $dir is never used again
```suggestion
    $: {
        const dir = direction || (RTL_LANGUAGES_CODE.includes($locale) ? 'rtl' : 'ltr');
        $isRtl = dir === 'rtl';
    }
```

packages/visualizations/src/components/Table/constants.ts
```diff
@@ -7,3 +7,24 @@ export const DATA_FORMAT = {
...
+    'ml',
+    'ms',
+    'ps',
+    'sd',
+    'so',
+    'tk',
+    'ug',
+    'ur',
+    'yi',
+] as string[];
```
etienneburdet, line 30: as const probably (it's already string[]). I didn't find a good way to fetch this list from somewhere reliable. At the very list we want a source link as comment.

packages/visualizations/src/components/Table/store.ts
```diff
@@ -1,7 +1,10 @@
...
-import { writable } from 'svelte/store';
+import { writable, get } from 'svelte/store';
 
 const defaultLocale = navigator.language;
+const defaultDirection = 'ltr';
 
 // eslint-disable-next-line import/prefer-default-export
 export const locale = writable<string>(defaultLocale);
+export const direction = writable<string>(defaultDirection);
+export const isRtl = writable<boolean>(get(direction) === 'rtl');
```
etienneburdet, line 10: Only one is needed. If we really need both, then `isRtl` is a `derived`. But we're not using `direction` anywhere.

etienneburdet (changes requested):
I might be missing some bits, but I think `<table dir={dir} />` does most of the heavy work, both for columns ordering, cell rendering, a11y etc. and that we might just want to force some cells to always left/right (probably numbers, url, geo)

cybervinvin:
After discussions, we decided to close this PR.
The table must use the reading direction of the page itself and the native behavior is sufficient.
New PR here: https://github.com/opendatasoft/ods-dataviz-sdk/pull/253

PR #251, opened 2024-06-26
fix: upgrade linked version

etienneburdet:
## Summary

The goal for this PR is to fix the dependency version of @opendatasoft/visualizations in visualizations-react that wasn't correctly updated due to a manual update of versions.

### DIFF
--- a/packages/visualizations-react/package.json
+++ b/packages/visualizations-react/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@opendatasoft/visualizations-react",
-    "version": "0.25.0",
+    "version": "0.24.1",
     "license": "MIT",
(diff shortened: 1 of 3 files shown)


packages/visualizations-react/package.json
```diff
@@ -52,7 +52,7 @@
         "arrowParens": "avoid"
     },
     "dependencies": {
-        "@opendatasoft/visualizations": "0.24.1-beta.2",
+        "@opendatasoft/visualizations": "^0.24.1",
```
KevinFabre-ods, line 55:
```suggestion
        "@opendatasoft/visualizations": "0.24.1",
```

If I remember,  we had issue without the caret

PR #252, opened 2024-07-01
[FEAT] New table column type: media

KevinFabre-ods:
## Summary

The goal for this PR is to add a new table column type: `media`

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-46281](https://app.shortcut.com/opendatasoft/story/46281).

### Changes

Based on what has been initiated [here](https://github.com/opendatasoft/ods-dataviz-sdk/pull/245), we use tippy to render a tooltip that contain an image. 
Image are lazy loaded, which means that the browser request images only when the `img` tag is visible to the user.

## Review checklist

- [x] Description is complete


KevinFabre-ods:
👋  @etienneburdet and @cybervinvin 

With our last PRs being merged, this PR is now ready to be reviewed.

packages/visualizations/src/components/Table/Cell/Format/ImageFormat.svelte
```diff
@@ -0,0 +1,79 @@
...
+            <div class="ods-visualization-table__image-error-container">
+                {@html errorContent(rawValue)}
+            </div>
+        {:else}
+            <div class="ods-visualization-table__image-loading-container">
+                {@html loadingContent(rawValue)}
+            </div>
+        {/if}
+    </div>
+</div>
```
etienneburdet, line 63: Probably worth reworking the calss as we suggested. The path to override shoud be something like "ods-viz-table ods-viz-image-error-container".

packages/visualizations/src/components/Table/Cell/Format/ImageFormat.svelte
```diff
@@ -0,0 +1,79 @@
...
+            {@html image?.outerHTML}
+        {:else if hasError}
+            <div class="ods-visualization-table__image-error-container">
+                {@html errorContent(rawValue)}
+            </div>
+        {:else}
+            <div class="ods-visualization-table__image-loading-container">
+                {@html loadingContent(rawValue)}
+            </div>
+        {/if}
```
etienneburdet, line 61: I don't think we should use `@html` to replace the lack of children in the SDK for now. It opens up a lot of security breaches (not for `new Image` which is safe) and forces the user to write HTML in template strings.
IMO, we should have something like this:

```suggestion
        <img 
            on:load={() => { isLoading = false; tippyInstance.setContent(...)}}
            on:error={(e) => { error = e; }}
            src={src}
            alt={alt}
        />
        {#if error}
            <div class="ods-visualization-table__image-error-container">
                {error}
            </div>
        {/if}
        {#if isLoading}
            <div class="ods-visualization-table__image-loading-container">
                maybe some default loader
            </div>
        {/if}
```
This leaves dom node creation to Svelte and limits errors to text and loader to pure CSS. These are limitations we should accept for now IMO.

packages/visualizations/src/components/Table/types.ts
```diff
@@ -95,14 +95,28 @@ export type GeoColumn = BaseColumn & {
...
 };
 
+/**
+ * Render an interactive text that displays an image in a tooltip on tab/hover.
+ */
+export type ImageColumn = BaseColumn & {
+    dataFormat: typeof DATA_FORMAT.image;
+    options?: {
+        display?: (v: string) => string;
+        alt?: (v: string) => string;
```
etienneburdet, line 105: Do we need a callback for alt? I don't see alt being built from URL anyways. Mabybe from another column, but practical case will most likely be `() => 'my alt'`, or `() => someVarFromElseWhere`, no?

etienneburdet (commented):
Looking mostly good and working well. IMO we can leverage image events a bit more, which would simplify the code a bit.

packages/visualizations/src/components/Table/types.ts
KevinFabre-ods: Indeed not very useful.
Maybe when a cell has access to all values for a row we can expose an option that takes these values as args

packages/visualizations/src/components/Table/Cell/Format/ImageFormat.svelte
KevinFabre-ods: Users of this repository are developers. We are not responsible for any security breaches that may arise from the values displayed in the table. However, components APIs should be clear enough for developers to understand when they will directly change the DOM with an option.

We have a similar mechanism in place when displaying content in the MapWebGL popup (see: [GitHub link](https://github.com/opendatasoft/ods-dataviz-sdk/blob/2f2b66328c3de126055057a95dd42de816b4c220/packages/visualizations/src/components/Map/WebGl/WebGl.ts#L351)).


I still simplify the options to only allow customization of the loading and error messages.

packages/visualizations/src/components/Table/Cell/Format/ImageFormat.svelte
KevinFabre-ods: Done https://github.com/opendatasoft/ods-dataviz-sdk/pull/252/commits/17c572def06df5b14d2c944617dfc942f6822cc0

KevinFabre-ods:
@etienneburdet and @cybervinvin

Let's manage our expectations for this PR. I'll reopen it when we have the time to work on new features for the table

PR #254, opened 2024-07-03
Chore: add path aliases to svelte

etienneburdet:
## Summary

The goal for this PR is to add path aliases to svelte. This requires to upgrade a number of critical packages.

### Changes
So the main thing is: do not use alisases in the entrypoint so that TS can output its index.d.ts safely. With that done, we can use aliases everywhere else, rollup will take care of aliases. This is no replacement for a real flat .d.ts file, but still works.

I also moved a few types to the root `types` instead of `components/types` both files seemed to serve the same purpose. The second thing is: we might actually want to have the `type.ts` being our public interface, with all exported handwritten there (and reimported along internal types in components).

I upgraded rollup and TS in the process, since the original plan was to use some rollup plugin to the aliasing for us.

Upgrade to svelte 4 still breaks in some places.

## Open discussions
For future releases, we might want to either : maintain a handwritten index.d.ts with our public interface (and import it back into components), or use something like API Extractor and mark some types as @internal.


packages/visualizations/rollup.config.mjs
```diff
@@ -1,28 +1,43 @@
...
 import commonjs from '@rollup/plugin-commonjs';
 import json from '@rollup/plugin-json';
 import { babel } from '@rollup/plugin-babel';
 import replace from '@rollup/plugin-replace';
 import { defineConfig } from 'rollup';
-import pkg from './package.json';
+import pkg from './package.json' with { type: 'json' };
 
 const production = !process.env.ROLLUP_WATCH;
+const __fileName = fileURLToPath(import.meta.url);
```
etienneburdet, line 20: __dirName is not available in module context

KevinFabre-ods (approved):
🚀 

Thank you for upgrading all those packages 🙏

PR #255, opened 2024-07-08
Chore/update actions

KevinFabre-ods:
## Summary

The goal for this PR is to update version of github actions we used in our CI. 
Currently we have warnings about incompatible Node version requirement 

**Before**
image

**After**

image

--- 

image

Also
- [removed a table Story from Chromatic](https://github.com/opendatasoft/ods-dataviz-sdk/pull/255/commits/9cf57981dfcad38fdc6e5a0f6db340256fb9722f) as the RTL addon we just added doesn't apply on the chromatic snapshot.
- publish Chromatic snapshots only with Node 20

### DIFF
--- a/.github/workflows/api-client.yml
+++ b/.github/workflows/api-client.yml
@@ -38,7 +38,7 @@ jobs:
         working-directory: ./packages/api-client
 
       - name: Comment
-        uses: romeovs/lcov-reporter-action@v0.4
+        uses: romeovs/lcov-reporter-action@v0.4.0
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           lcov-file: ./packages/api-client/coverage/lcov.info
--- a/.github/workflows/api-client.yml
+++ b/.github/workflows/api-client.yml
@@ -8,22 +8,17 @@ on:
       - main
 jobs:
   build:
-    name: Get api client test coverage on node ${{ matrix.node }} and ${{ matrix.os }}
-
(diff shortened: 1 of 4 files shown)


KevinFabre-ods:
@etienneburdet Should we remove the Node matrix for the `api-client` action?
What the point of having twice (https://github.com/opendatasoft/ods-dataviz-sdk/pull/255#issuecomment-2217064149 & https://github.com/opendatasoft/ods-dataviz-sdk/pull/255#issuecomment-2217064636) the coverage message?

etienneburdet (approved):
LGTM, that was needed indeed!

I'm OK with moving to node 20.x rather than matrix. As for the code coverage, at the very least I would remove the matrix indeed. Why not removing it all together: we're not really using it. Or if we keep it, actually use it accross all the repo and our workflow.

KevinFabre-ods:
> LGTM, that was needed indeed!
> 
> I'm OK with moving to node 20.x rather than matrix. As for the code coverage, at the very least I would remove the matrix indeed. Why not removing it all together: we're not really using it. Or if we keep it, actually use it accross all the repo and our workflow.

`api-client` package is very independent of the two others. Why not keep it, it can help to check that new code has a code coverage. But, it's true that we not are really using it. 

Having the same mechanism for the visualization packages can't work.
Code coverage for `visualizations-react` would means nothing as all logic is in `visualizations` (unless there is a way to get code coverage of a dependency)

PR #256, opened 2024-07-12
feat(Table) add `size` option on cells

cybervinvin:
## Summary
[explo] The goal for this PR is to add an option (size) to define the width of columns.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-48043](https://app.shortcut.com/opendatasoft/story/48043/explo-sdk-table-width-options-for-a-column).

### Changes

New `size` option for cell with predefined values (`small`, `medium`, `large` and `fit-content` [default])

#### Breaking Changes

none

### Changelog

In the Table component: It's now possible to define a size for the column width

## Open discussion

explo ... all comments are welcome

## To be tested

in storybook

## Review checklist

- [X] Description is complete


etienneburdet (commented):
LGTM. 

I wonderedl for a while if it was not CSS customization realm, but without this option it would hard to have to columns of the same type having different width. 

We could argue that columns of the same type should have the same width though—in which case the type class is sufficient as a selector. But overall, I don't see any drawback to adding it, it's optional anyways if someone doesn't want to use it.

packages/visualizations/src/components/Table/Cell/Cell.svelte
```diff
@@ -24,23 +26,37 @@
...
     :global(.ods-dataviz--default td.table-data--url) {
         text-overflow: ellipsis;
         overflow: hidden;
-        width: max-content;
         min-width: 40px;
-        max-width: 240px;
     }
-    :global(.ods-dataviz--default td.table-data--long-text > span) {
+    :global(.ods-dataviz--default td.table-data--long-text > span),
+    :global(.ods-dataviz--default td.table-data--geo > div) {
```
KevinFabre-ods, line 37: image

Geo labels are no longer vertically centered

Are we sure we want to display geo labels on 3 lines?

PR #258, opened 2024-07-18
[STYLE][TABLE]: "no records" label is always visible

KevinFabre-ods:
## Summary

The goal of this PR is to fix an issue where "no records" is not visible in a large table with a scroll section. Currently, having "no results" renders the table with a 'void' row, making it appear broken.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-51257](https://app.shortcut.com/opendatasoft/story/51257).

### Changes
Update the Empty part

## Review checklist

- [x] Description is complete

### DIFF
--- a/packages/visualizations/src/components/Table/EmptyRow.svelte
+++ b/packages/visualizations/src/components/Table/EmptyRow.svelte
@@ -3,9 +3,9 @@
 </script>
 
 <tr>
-    <div>
+    <td>
         <em>{label}</em>
-    </div>
+    </td>
 </tr>
 
 <style lang="scss">
@@ -13,7 +13,8 @@
     tr {
         height: $row-height;
     }
-    div {
+    td {
+        box-sizing: border-box;
         position: absolute;
         left: 0;
         right: 0;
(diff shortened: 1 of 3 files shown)


packages/visualizations/src/components/Table/EmptyRow.svelte
```diff
@@ -0,0 +1,25 @@
+<script lang="ts">
+    export let label = 'No records found...';
+</script>
+
+<tr>
+    <div>
+        <em>{label}</em>
+    </div>
+</tr>
```
cybervinvin, line 9: `div` can't be inside a `tr`. I think it's incorrect for HTML.

cybervinvin (changes requested):
I think there is an error in `EmptyRow` (if it doesn't I will be shocked 😲 ).

packages/visualizations/src/components/Table/EmptyRow.svelte
KevinFabre-ods: Indeed, you're right! Fixed in https://github.com/opendatasoft/ods-dataviz-sdk/pull/258/commits/e9dacf7d3cd574e9d0d8fde0661f8f1e03ee8d36

I struggled a bit to do what I wanted with a `td` that why I used a `div`.
But everything looks fine with a `td`, I guessed my CSS wasn't good enough at first.

cybervinvin (approved):
All is good 👍🏼 
go merge 🚀

PR #259, opened 2024-08-30
[API-CLIENT][FEAT] Update and add new dataset types

KevinFabre-ods:
## Summary

The goal for this PR is to update some dataset types

I need keys in the Dataset type to develop the new export page.

### Changes

The main update involves the OdsDataset interface, where we’ve removed the optional keys. Having a 'partial' object didn't make sense. For more details, refer to [this Slack message](https://opendatasoft.slack.com/archives/G019JBY7RJN/p1723549652625339).

Additionally, I introduced new sub-types for OdsDataset, such as OdsDatasetAlternativeExport, OdsDatasetAttachment, and OdsDatasetField.

I also created a new constants file to expose values like ODS_DATASET_FIELD_TYPE.

- [x] Description is complete

### DIFF
--- a/packages/api-client/src/client/types.ts
+++ b/packages/api-client/src/client/types.ts
@@ -1,5 +1,6 @@
 // Type hints for Api response
 
+import { ValueOf } from "../utils";
 import { EXPORT_CATALOG_FORMAT, EXPORT_DATASET_FORMAT, ODS_DATASET_FIELD_TYPE } from "./constants";
 
 export interface Facet {
@@ -20,7 +21,7 @@ export interface Link {
     rel: string;
 }
 
-export type OdsDatasetFieldType =  typeof ODS_DATASET_FIELD_TYPE[keyof typeof ODS_DATASET_FIELD_TYPE];
+export type OdsDatasetFieldType =  ValueOf<typeof ODS_DATASET_FIELD_TYPE>;
 
 export interface OdsDatasetAttachement {
     id: string;
@@ -90,10 +91,8 @@ export interface ApiExport<T> {
 }
 
 
-export type ExportCatalogFormat =
-    typeof EXPORT_CATALOG_FORMAT[keyof typeof EXPORT_CATALOG_FORMAT];
+export type ExportCatalogFormat = ValueOf<typeof EXPORT_CATALOG_FORMAT>;
 
 
 
-export type ExportDatasetFormat =
-    typeof EXPORT_DATASET_FORMAT[keyof typeof EXPORT_DATASET_FORMAT];
+export type ExportDatasetFormat = ValueOf<typeof EXPORT_DATASET_FORMAT>;
--- a/packages/api-client/src/client/types.ts
+++ b/packages/api-client/src/client/types.ts
@@ -1,5 +1,7 @@
 // Type hints for Api response
 
+import { EXPORT_CATALOG_FORMAT, EXPORT_DATASET_FORMAT, ODS_DATASET_FIELD_TYPE } from "./constants";
+
 export interface Facet {
     name: string;
     count: number;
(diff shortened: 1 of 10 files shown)


KevinFabre-ods:
@5k4nd Can you please tell me why I see difference of keys between the document and actual data return by the API when catalog datasets?

- `links` is in the documentation but not in data
- `alternative_exports` is not in the documentation but is in data

packages/api-client/src/client/types.ts
```diff
@@ -18,22 +20,40 @@ export interface Link {
     rel: string;
 }
 
-export interface OdsDataset {
-    dataset_id?: string;
-    dataset_uid?: string;
-    has_records?: boolean;
-    data_visible?: boolean;
-    features?: string[];
+export type OdsDatasetFieldType =  typeof ODS_DATASET_FIELD_TYPE[keyof typeof ODS_DATASET_FIELD_TYPE];
```
etienneburdet, line 23: I added the utility type `ValueOf<>` on the platform, we can maybe add it here. (or I could make an npm package…). https://github.com/opendatasoft/platform/blob/8497ce8419cacc607c41b79ff0b87534524dddf9/ods/react/src/utils/typeScript.ts#L29

etienneburdet (approved):
Oui!

As said, some sort of automation would be great, espcially on those times where lots of things are changing. But that's blocking for now. I'll try to look into it though one day…

packages/api-client/src/client/types.ts
KevinFabre-ods: > or I could make an npm package…
Yes, exactly, like a TypeScript helpers library. Because it would feel wrong to import in our project ts helpers from this lib

But should this library be under this repository?

KevinFabre-ods: Ok so I start simply with a utils folder
https://github.com/opendatasoft/ods-dataviz-sdk/pull/259/commits/f5f1ba215f12bf31a7279995b1a7c70b5b684259

5k4nd:
> @5k4nd Can you please tell me why I see difference of keys between the documentation and actual data returned by the API when I fetched catalog datasets?
> 
>     * `links` is in the documentation but not in data
> 
>     * `alternative_exports` is not in the documentation but is in data

Hello @KevinFabre-ods, the doc may not be up-to-date indeed.
The HATEOAS links are now hidden by default since the Explore API v2.1, and you can make them appear using the `include_links` REST param ([e.g. API call](https://public.opendatasoft.com/api/explore/v2.1/catalog/datasets/?include_links)). Links will be available in a separate `_links` object in the JSON root, and all objects will also have a `_links` sub-object.

As for `alternative_exports`, _it is_ actually in the data, maybe your're missing something? [e.g. API call](https://public.opendatasoft.com/api/explore/v2.1/catalog/datasets/arbresremarquablesparis2011?select=title,datasetid,alternative_exports)

KevinFabre-ods:
@5k4nd 



Thanks for the explanation about the links, I'll update the code accordingly.

> As for `alternative_exports`, _it is_ actually in the data, maybe your're missing something? [e.g. API call](https://public.opendatasoft.com/api/explore/v2.1/catalog/datasets/arbresremarquablesparis2011?select=title,datasetid,alternative_exports)

Yes, I said it was in the data but can't find it in the documentation

5k4nd:
> Yes, I said it was in the data but can't find it in the documentation

Ah, sorry, I take the point for the doc then, thanks 👍

KevinFabre-ods:
@etienneburdet I made some change [here](https://github.com/opendatasoft/ods-dataviz-sdk/pull/259/commits/716cbaedb462a813eab6f19959a87a9247e40cf5) to add the optional `_links` key. 

Also I changed a bit of naming and updated types:
- `ApiDatasets` renamed into `ApiCatalog`
- `OdsRecord` renamed into `OdsDatasetRecord`
- `ApiRecords` renamed into `ApiDatasetRecords`

`ApiCatalog` and `ApiDatasetRecords` now use the `ApiQuery` type under the hood as they are a standard way of fetching data thought the generic query system. 

Impact:

For our platform, only the `ApiQuery` type is used. Impact should be limited and some of them could be replace by `ApiCatalog`

However I don't know how these types are used in our client applications (cc @fpassaniti ). We should make available a release not to explain the change.

Also, I suggest to release a beta version to test this change first in Platform.

@etienneburdet What do you think

fpassaniti:
@KevinFabre-ods here are the usages for Alim'Confiance 

image

And for Frais et local 

image

KevinFabre-ods:
> @KevinFabre-ods here are the usages for Alim'Confiance
> 
> image
> 
> And for Frais et local
> 
> image
> 
> If I understand correctly, it should be ok for us

Ok thanks, so you don't import api-client TS types, so these projects should be fine

packages/api-client/src/client/types.ts
```diff
@@ -68,14 +67,15 @@ export interface ApiFacets {
...
-    total_count: number;
+export type ApiQuery<T extends object> = OdsDataWithLinks & {
+    total_count?: number;
     results: T[];
 };
 
-export type ApiCatalog = ApiQuery<OdsDataset>;
-export type ApiDatasetRecords<T extends object> = ApiQuery<OdsDatasetRecord<T>>;
-
+export type ApiCatalog = ApiQuery<OdsDataset> & { total_count: number };
```
etienneburdet, line 75: Maybe do something like:
```
export type ApiCatalog<Shape = OdsDataset> = ApiQuery<Shape> & { total_count: number };
```
To deal with selects?

etienneburdet (commented):
Mostly OK, I just made some opinonated comment that could go both ways. I think we can improve a bit the "select" case as discussed.

packages/api-client/src/client/types.ts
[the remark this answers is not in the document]
KevinFabre-ods: Indeed it's redundant

KevinFabre-ods:
@etienneburdet your last round of comments have been taken in consideration in https://github.com/opendatasoft/ods-dataviz-sdk/pull/259/commits/66f4cf29f2488e9855f82bec1f0d9657fed285d8

etienneburdet:
> @etienneburdet your last round of comments have been taken in consideration in [66f4cf2](https://github.com/opendatasoft/ods-dataviz-sdk/commit/66f4cf29f2488e9855f82bec1f0d9657fed285d8)

All good for me 🚀

PR #260, opened 2024-09-19
[FIX][API-CLIENT] Update dataset exports formats

KevinFabre-ods:
## Summary

The goal for this PR is to add new dataset export formats

- geographic formats: fgb, gpx and parquet
- calendar format: ical

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-51661](https://app.shortcut.com/opendatasoft/story/51661).


richterb (approved):
I'm not sure why the older one such as GPX and iCal were left out, they're used in ODS UIs so they're not secret on purpose. Probably just an oversight

PR #263, opened 2024-10-29
Url format fallback to short text

etienneburdet:
## Summary

The goal for this PR is to add a fallback to URL format so that text can be autoformatted as url if an url is detected.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-53035](https://app.shortcut.com/opendatasoft/story/sc-53035).

### Changes
I mainly changed the rendered to short text instead rawvalue in case it's not an URL, so that other checks and formatting could run. I sligthly changed the warning the too.

The other solution could have been to add an `autoFormatUrl` option to short-text that rendered us URL. In both cases we check all entries anyways, so makes for one less option.

## Review checklist

- [x] Description is complete

### DIFF
--- a/packages/api-client/src/index.ts
+++ b/packages/api-client/src/index.ts
@@ -4,8 +4,8 @@
  */
 
 export * from './client';
+export * from './client/constants';
 export * from './client/error';
 export * from './client/types';
-export * from './client/constants';
 export * from './odsql';
 export * from './odsql/clauses';
--- a/packages/api-client/src/index.ts
+++ b/packages/api-client/src/index.ts
@@ -4,8 +4,8 @@
  */
 
 export * from './client';
-export * from './client/constants';
 export * from './client/error';
 export * from './client/types';
+export * from './client/constants';
 export * from './odsql';
 export * from './odsql/clauses';
--- a/packages/api-client/test/client.test.ts
+++ b/packages/api-client/test/client.test.ts
@@ -2,8 +2,8 @@
  * @jest-environment jsdom
  */
 
+import { beforeEach, describe, expect, it } from '@jest/globals';
 import { enableFetchMocks } from 'jest-fetch-mock';
-import { expect, it, beforeEach, describe } from '@jest/globals';
 import { ApiClient, field, fromCatalog } from '../src';
 import { AuthenticationError, NotFoundError, ServerError, UserError } from '../src/client/error';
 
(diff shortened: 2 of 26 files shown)


KevinFabre-ods:
@etienneburdet 👋 

Just out of curiosity, looking at the diff, do you have a extension in your IDE that auto sort imports?
Have all files been sorted? It seems that imports in `visualizations-react` aren't sorted 🤔

KevinFabre-ods (commented):
Nice addition this fallback 👌 

I just have a remark about the usage of `display` in the format function

KevinFabre-ods:
> > @etienneburdet 👋
> > Just out of curiosity, looking at the diff, do you have a extension in your IDE that auto sort imports? Have all files been sorted? It seems that imports in `visualizations-react` aren't sorted 🤔
> 
> I had a lint problem that surprised me, ran a `prettier --fix` and it did this. It's only on changed directiores, which might explain why `visualization-react` was skipped. I don't why we didn't have that before. We didn't upgrade prettier and I didn't run that through my IDE.

Could you maybe extract it in a specific commit (like chore: ...) to avoid the auto versioning for api-client by Lerna when we'll do a release?

etienneburdet:
I removed all prettier diff, I'll make a dedicated PR to avoid that in future PR. I don't think tha's Lerna upgrade causing that though… 😕

KevinFabre-ods:
> I don't think tha's Lerna upgrade causing that though… 😕

That's not what I meant. 
My point was that having a conventional commit (like https://github.com/opendatasoft/ods-dataviz-sdk/pull/263/commits/5972949823d2e68fc8f2b09d15f2b674965930d2)  with changes in the `api-client` package would trigger an upgrade for the next release process (made by Lerna). We don't want to upgrade when files are only being reformatted.

KevinFabre-ods (approved):
Nice improvement, maybe we could update a bit the data for our table story to have one row where the link is broken.So that snapshots would capture the fallback to the short-text

etienneburdet:
> > I don't think tha's Lerna upgrade causing that though… 😕
> 
> That's not what I meant. My point was that having a conventional commit (like [5972949](https://github.com/opendatasoft/ods-dataviz-sdk/commit/5972949823d2e68fc8f2b09d15f2b674965930d2)) with changes in the `api-client` package would trigger an upgrade for the next release process (made by Lerna). We don't want to upgrade when files are only being reformatted.

No sorry, I meant to search what is causing `prettier --fix` to sort alphabetically all of a sudden and only thing I see changing an NX version. But well, we'll do a separate run to avoid having that in all PR, I totally see the point 👌

PR #265, opened 2024-11-12
feat: add clickable and hoverable rows to table

etienneburdet:
## Summary

The goal for this PR is to add clickable and hoverable rows to the table.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-53066](https://app.shortcut.com/opendatasoft/story/53066).

### Changes
I added a `rows` options to add hoverable and clickable. The click is implement as a button on the first column so that it doesn't interact with other clickable cells (links, geo etc.).

The SVG is hardcoded for now since we don't support slots yet.

## Review checklist

- [x] Description is complete


KevinFabre-ods (changes requested):
- In the new Storybook story, I wasn't able to click on a record when navigating the table using only my keyboard. 
One of Benoit R’s comments on the Shortcut ticket mentions this requirement. He suggested we might not need it for the first implementation. However, since you didn’t make any comments (in the code or in the PR description) about it, I wanted to bring up this issue.

- Currently, I can only trigger the click event when clicking on the icon button. Shouldn’t the entire row be clickable? For tables with many columns, if I scroll to the end of the table, I then have to scroll back to the beginning to click on the icon button. This feels like bad UX. 
However, the mockup doesn’t provide any information about this. One possible improvement could be to make the column sticky.

packages/visualizations/src/components/Table/Row.svelte
```diff
@@ -0,0 +1,83 @@
...
+    };
+
+    $: onClick = () => {
+        if (rows?.onClick) {
+            rows.onClick(record);
+        }
+    };
+</script>
+
+<tr on:mouseenter={onMouseEnter} on:mouseleave={onMouseLeave}>
```
richterb, line 32: It's a micro optimization, but ideally we shouldn't have these handlers attached at all if there's no `rows?.onClick`.

richterb:
> I'll test for the keyboard navigation indeed, it should be achievable with a button (most likely the visibility attribute issue)

I suspect the only way to achieve this would be to use the usual CSS weird hack, you can find it in the platform, look for the `screenReaderOnly` class. It makes the element invisible BUT still rendered in the page flow so focusable.

packages/visualizations/src/components/Table/Row.svelte
```diff
@@ -0,0 +1,83 @@
...
+    {#if rows?.onClick}
+        <td class="button-cell">
+            <button on:click={onClick} class:visible={isRowHovered}>
```
richterb, line 35: There's no text in the button so it probably needs an `aria-label`. I'm not sure there's something generic enough, maybe it's worth an exposed property (here we'd pass "Open record" I guess).

richterb (commented):
A few small details, but I agree with Kevin's point regarding losing the button when scrolling horizontally. This is something that can be discussed with the design team, and if the answer is to have a sticky column, I suppose we can do this a bit later in a separate PR.

etienneburdet:
So in order to format records in the side panel (or anywhere else), I went with exporting the formatters. We have one case, there might quite some others. To do so I had to rewrite our table typing a little bit. Especially I had to change the `<svelte:component …` in Cell, so that not only we pass correctly typed props, but we can also add custom or fixed props.

The option eventhandler attach is… a bit weird. I've never seen this pattern actually (in React neither I think), but well it works. Maybe we should just pass `null` 🤷 . Else we could have some <RowWithEvent> component attaching stuff conditionnally, but since every props is optional, there is a bit of combinatory.

packages/visualizations/src/index.ts
```diff
@@ -21,5 +21,15 @@ export * from './components/MarkdownText/types';
...
+export {
+    BooleanFormat,
+    DateFormat,
+    GeoFormat,
+    ShortTextFormat,
+    LongTextFormat,
+    NumberFormat,
+    URLFormat,
+} from 'components/Format';
+export * from 'components/Format/types';
```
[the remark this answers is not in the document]
richterb: Either way the code should be the same for every export, but I'm wondering if this is something that works for you and not for me, and what would be special about my setup...

packages/visualizations/src/components/Table/Row.svelte
```diff
@@ -0,0 +1,122 @@
...
+    };
+    $: onMouseleave = () => {
+        if (rows?.onMouseLeave) {
+            rows.onMouseLeave(record);
+        }
+        isRowHovered = false;
+    };
+
+    $: if (rowElement) {
+        if (rows) {
```
KevinFabre-ods, line 26: IMO, the `rows` naming find its limit in this file. It's weird to have a plural form of the word in the `Row` component.

We are starting to have a lot of interactions for this component (sort and now row events). Maybe we could starting thinking of other system to propagate events

packages/visualizations/src/components/Table/Row.svelte
```diff
@@ -0,0 +1,122 @@
...
+        border-bottom: 1px solid var(--border-color);
+    }
+
+    :global(.ods-dataviz--default) tr:last-child {
+        border-bottom: none;
+    }
+
+    :global(.ods-dataviz--default) button {
+        background-color: transparent;
+        color: #142e7b;
```
KevinFabre-ods, line 82: this color should be customizable like we did for the sort icon

KevinFabre-ods (commented):
I let the UX review to the product/design team as you took the time to build a beta tag 👍 

Here my comments about the diff, nothing really block to merge the branch

packages/visualizations/src/components/Table/Row.svelte
KevinFabre-ods: Now the with the latest changes, even if `rowProps` is defined, we can't be sure that `onMouseEnter` is defined.

Complete check should be `rowProps?.onMouseEnter`.
But do we want do to it like this as `handleMouseEnter` is already making the check on `onMouseEnter`? 


Can we do something like this? 

```svelte
    $: handleMouseEnter = onMouseEnter ?  () => {
        onMouseEnter(record);
        isRowHovered = true;
    } : undefined;
    
    <tr on:mouseenter={handleMouseEnter} ... />
 ```

**Review by KevinFabre-ods — COMMENTED** [review]
Since this PR will export the format components, I pointed out a different approach to handling incorrect raw values. This could be a dedicated story to avoid adding more things to review here.

Also, I couldn’t see the icon when the button is focused.

File: packages/visualizations/src/components/Format/LongTextFormat.svelte [hunk 6]
```diff
@@ -0,0 +1,24 @@
...
+        if (typeof v !== 'string') {
+            warn(v, 'text');
+        }
+        if (display) {
+            return display(v as string);
+        }
+        return v;
+    };
+
+    $: value = format(rawValue);
```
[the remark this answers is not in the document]
etienneburdet: I think https://github.com/opendatasoft/ods-dataviz-sdk/pull/267 is the right place to implement that, since I generalize formatters there. But agreed on the general principle of inversion indeed 👌

packages/visualizations/src/components/Table/Row.svelte
```diff
@@ -0,0 +1,100 @@
...
+<tr on:mouseenter={rowProps && handleMouseEnter} on:mouseleave={rowProps && handleMouseLeave}>
+    {#if rowProps?.onClick}
+        <td class="button-cell">
+            <button
+                on:click={rowProps && handleClick}
+                aria-label={actionAriaLabel || 'Expand Record'}
+            >
+                <span class:visually-hidden={!isRowHovered}>
+                    <ZoomIcon />
+                </span>
```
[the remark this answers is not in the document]
etienneburdet: ha yes forgot to add back focusin et focusout 👍

etienneburdet:
> Can we do something like this?
> 
>     $: handleMouseEnter = onMouseEnter ?  () => {
>         onMouseEnter(record);
>         isRowHovered = true;
>     } : undefined;
>     
>     <tr on:mouseenter={handleMouseEnter} ... />

I thougth the same at first, but we can't because we need  to set the hovered state to true false for the `on:click` even if no `onMouseEnter/Leave` is specified.

KevinFabre-ods:
> > Can we do something like this?
> > ```
> > $: handleMouseEnter = onMouseEnter ?  () => {
> >     onMouseEnter(record);
> >     isRowHovered = true;
> > } : undefined;
> > 
> > <tr on:mouseenter={handleMouseEnter} ... />
> > ```
> 
> I thougth the same at first, but we can't because we need to set the hovered state to true false for the `on:click` even if no `onMouseEnter/Leave` is specified.

Oh yes! indeed

KevinFabre-ods (approved):
LGTM ! ✅ 

Will you make a feature upgrade? If yes, the packages files needs to be cleaned to remove references about a 0.27.1 fix

packages/visualizations/src/components/Table/Row.svelte
```diff
@@ -0,0 +1,105 @@
...
+    on:focusin={rowProps && handleMouseEnter}
+    on:focusout={rowProps && handleMouseLeave}
+>
+    {#if rowProps?.onClick}
+        <td class="button-cell">
+            <button
+                on:click={rowProps && handleClick}
+                aria-label={actionAriaLabel || 'Expand Record'}
+            >
+                <span class:visually-hidden={!isRowHovered}>
```
richterb, line 43: The mouseenter/mouseleave can be tricky and skipped in some cases. In our use case with a modal I think mouseleave is never triggered.

https://github.com/user-attachments/assets/1197ab1d-287e-4719-8594-ab1266c1a9f5

Maybe this can be achieved instead by a regular `tr:hover ...` in CSS to visually hide the button? (which as far as I know handles that case well)

etienneburdet: ha, I get it it's the focusin/out that does that. It worked without it but then we had no keyboard and now I brought it back we don't lose focus when opening the modal. I'll check if it's fixable only on the platform side.

richterb (approved):
🚀 thanks

PR #266, opened 2024-11-26
Feature/sc-48047/table-sdk-sticky-columns

etienneburdet:
## Summary

The goal for this PR is to add sticky columns to the table. I tried to support a "pin" column scenario, where users can add/remove multiple columns to sticky side.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-48047](https://app.shortcut.com/opendatasoft/story/48047).

## Changes
**Sticky style**
I chose `position: sticky` with computed offsets. 
Other possibilities: back to full div, second non scrolling table.
Pros: a11y because it's a single table.
Cons: shadow is a bit harder to do, requires js to compute offsets.

**Sticky state API**
I chose to declare the sticky in the column declaration.
Other possibilities: declaring a `stickyColumns: …` in the config, exposing an headless state manager (`column.pin()`) à la/with Tanstack table.
Pros: can only pin defined colmun no id mismatch.
Cons: can be a it of a burden to update since it's an object, start mixing definition and state management.

**Sticky state internal**
I went with initilazing a Map and reseting both the map and scroll on each change of `columns` (even non-sticky) and sorting columns based on sticky and their initial order.
Other possibilities: using array/indexes, pushing in order of selection, not messing with order (sticky only when "met"), using `column.sticky` in palce.
Pros: no need for index matching in 3 different places, everything derived from base store, seperation between user config and internal state.
Cons: May reset on colmuns update that didn't require it.

I made a specific `Th` component first because it was getting really big, but mostly to avoid an array of `clientWidth` since array reactivity in Svelte can be annoying.

The shadow is made with a linear gradient, because box-shadow cannot be applied to only one side and thus required some keyed z-index manipulations for each column to hide the non-wanted parts.

### Breaking changes
Since I needed border: separate, we end-up with a 1px diffs an all table stories, one way or another. Here I removed the extra 1px td padding, the other way around added it. I thought it wasn't worth the 0.5px padding just to minimize the diff.

### Changelog
 "It's now possible to add sticky columns to the table, that remains visible while scrolling".

## Review checklist

- [x] Description is complete


packages/visualizations/src/components/Table/Cell/Cell.svelte
```diff
@@ -1,46 +1,66 @@
...
+        {/if}
+    </div>
 </td>
 
-<style>
+<style lang="scss">
+    @import '../sticky';
     :global(.ods-dataviz--default td) {
-        padding: var(--spacing-75);
+        background-color: white;
```
KevinFabre-ods, line 30: image

We lost the ability to apply a background color on tbody, tr

Maybe we could move this style to the tr tag?

packages/visualizations/src/components/Table/Headers/Headers.svelte
```diff
@@ -1,14 +1,16 @@
...
+    export let isHorizontallyScrolled: boolean;
     export let columns: Column[];
 </script>
 
 <thead>
     <tr>
-        {#each columns as column}
-            <th class={`table-header--${column.dataFormat}`}>
+        {#each columns as column (column.key)}
+            <Th dataFormat={column.dataFormat} key={column.key} {isHorizontallyScrolled}>
```
KevinFabre-ods, line 13: Do you think we could put L14-L24 into the `Th` component ?

packages/visualizations/src/components/Table/Headers/Th.svelte
```diff
@@ -0,0 +1,39 @@
...
+        stickyColumnsWidth.updateColumn(key, clientWidth);
+    }
+</script>
+
+<th
+    style={`--sticky-offset: ${$stickyColumnsOffset.get(key)}px`}
+    class={`table-header--${dataFormat}`}
+    class:sticky={$stickyColumnsWidth.has(key)}
+    class:isLastSticky={key === $lastStickyColumn}
+    class:isHorizontallyScrolled
```
KevinFabre-ods, line 21: These three classes are shared between `Th` and `Cell`
What do you think to create a hook to build these sticky classes ?

packages/visualizations/src/components/Table/Table.svelte
```diff
@@ -12,12 +13,44 @@
...
+            }
+            return colA?.sticky ? -1 : 1;
+        });
+        stickyColumnsWidth.reset();
+        sortedStickyColumns
+            .filter((col) => col?.sticky)
+            .forEach((col) => {
+                stickyColumnsWidth.updateColumn(col.key, 0);
+            });
+    }
```
KevinFabre-ods, line 41:
```suggestion
    $: if (columns && scrollBox) {
        scrollBox.scrollLeft = 0;
        stickyColumnsWidth.reset();
        columns.forEach((col) => {
                col?.sticky && stickyColumnsWidth.updateColumn(col.key, 0);
            });
    }
```

Did I miss something?
Do we want to get update the store state with only columns keys that are sticky?
(While keeping the original order of the column)

KevinFabre-ods (changes requested):
That's a nice feature GG! 👏 

The offset computation implementation is not so big and is reactive to column widths !(column definition doesn't need a size)

> Sticky style
> I chose position: sticky with computed offsets.
> Other possibilities: back to full div, second non scrolling table.
> Pros: a11y because it's a single table.
> Cons: shadow is a bit harder to do, requires js to compute offsets.

I think it's looking nice! The offset computation is quite lite to be fair.

> Sticky state API
> I chose to declare the sticky in the column declaration.
> Other possibilities: declaring a stickyColumns: … in the config, exposing an headless state manager (column.pin()) à la/with Tanstack table.
> Pros: can only pin defined colmun no id mismatch.
> Cons: can be a it of a burden to update since it's an object, start mixing definition and state management.

It's a good start to stay consistence with our other declarations (sort system for example).
You're right to point out that it can be heavy to declare if we have UIs where sticky columns could be change frequently by end users.

> Sticky state internal
> I went with initilazing a Map and reseting both the map and scroll on each change of columns (even non-sticky) and sorting columns based on sticky and their initial order.
> Other possibilities: using array/indexes, pushing in order of selection, not messing with order (sticky only when "met"), using column.sticky in palce.
> Pros: no need for index matching in 3 different places, everything derived from base store, seperation between user config and internal state.
> Cons: May reset on colmuns update that didn't require it.

For a start, IMO it's acceptable that the scroll resets when columns option change. There is room for optimization


Other than that, i left some comment to fix minors issues

packages/visualizations/src/components/Table/Cell/Cell.svelte
etienneburdet: It has to be on the td because we will have overlapping tds with no background elese. I changed the customization stories, it seems acceptable.

packages/visualizations/src/components/Table/Table.svelte
etienneburdet: I opted to sort the sticky column first so that they are already in their sticky position. In the version you shared colmun will stick "as they come". I thought it was pretty common for "pinned" columns to be first. This could also be done by putting sticky columns in a different array (in which case we don't need to sort).

KevinFabre-ods:
```js
[...columns].sort((colA, colB) => {
            if (Boolean(colA?.sticky) === Boolean(colB?.sticky)) {
                return 0;
            }
            return colA?.sticky ? -1 : 1;
        });
```

```js
columns.filter((col) => col?.sticky)
```

seems to produce the same result.
I don't understand why we need this sort function

packages/visualizations/src/components/Table/Headers/Th.svelte
KevinFabre-ods, line 1: I have a weird behavior where sometime the first header cell is not sticky, its position style stays to `relative`

I wasn't able to understand a clear pattern 

https://github.com/user-attachments/assets/b6d49e5f-0176-49ae-ba4d-46153810fb8c

KevinFabre-ods (changes requested):
Mostly for comment are about style modification.
Also, during my tests, sometimes the first sticky header column lost its sticky style

packages/visualizations/src/components/Table/Row.svelte
```diff
@@ -52,10 +52,6 @@
 </tr>
 
 <style>
-    :global(.ods-dataviz--default tr) {
-        border-bottom: 1px solid var(--border-color);
-    }
-
```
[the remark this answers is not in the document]
etienneburdet: Yes, they would disappear with boder-collapse separate, the cell being sticky and having a background (the cell ends up "on top" of the border). Easy enough to eliminate that last border though 👌

packages/visualizations/src/components/Table/Headers/Th.svelte
etienneburdet: That's a tricky one… https://github.com/sveltejs/svelte/issues/4776 🤯

etienneburdet:
Okaaaay, could remove all the diffs… 😓

packages/visualizations-react/stories/Table/StickyColumns.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+const data: Async<TableData> = {
+    value,
+    loading: false,
+};
+
+const Template: ComponentStory<typeof Table> = args => {
+    const { data: templateData, options: templateOptions } = args;
+    const { columns } = templateOptions;
+    const [stickyColumns, setColumns] = useState(columns);
```
KevinFabre-ods, line 22: Would it be possible to initiate the stories with some sticky columns?
So we could have chromatic snapshots that assert the sticky style for columns

KevinFabre-ods (commented):
LGTM 🚀 

I suggest a small change about sticky stories for Chromatic to have snapshots where sticky style is visible

KevinFabre-ods (approved):

packages/visualizations/src/components/Table/Headers/Th.svelte
```diff
@@ -0,0 +1,53 @@
...
+        {column.title}
+    {/if}
+</th>
+
+<style lang="scss">
+    @import '../sticky';
+    :global(.ods-dataviz--default th) {
+        text-align: left;
+        padding: var(--spacing-75);
+        background-color: white;
```
richterb, line 46: That one wasn't there before and impacts the platform. Is that intended?

packages/visualizations/src/components/Table/utils.ts
```diff
@@ -0,0 +1,22 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { get } from 'svelte/store';
+import { isNil } from 'lodash';
+import type { DataFormat, Column, ColumnOfType } from './types';
+import { lastStickyColumn } from './store';
+
+export const getStickyClasses = (column: Column, scrolled: boolean) =>
```
richterb, line 7: I really don't get how this could happen, but once loaded in the platform, we get the following class attribute on `th`:
```
table-header-- undefined false false svelte-1quk48x
```
How this function could return `undefined false false` eludes me 😳

richterb (commented):
There's a glitch when I test in the Storybook: If the first column I set as sticky is the first column of the table, everything is fine. If the first column I set as sticky is not the first column of the table, then the header of the first sticky column doesn't stick:


https://github.com/user-attachments/assets/9f7780cc-d4ef-4457-a8d6-34d47a18956d

packages/visualizations/src/components/Table/utils.ts
etienneburdet: really seems like it was not on the last version (I just checked it again) 🤔

etienneburdet:
@KevinFabre-ods @richterb Sorry for the "one more feature" but I just made the inspect/expand button sticky (which was the point in the first place). This involved tweaking padding a bit between scrolled/not scrolled.

I also made the sticky story with sticky row and some scroll by default so that we have a decent snapshot 👌

packages/visualizations/src/components/Table/utils.ts
richterb: Weird, I'll try again and inspect my node_modules. There's so many layers of indirection 🙈

packages/visualizations/src/components/Table/Headers/Th.svelte
richterb: As discussed, will be part of the PR that adapts the changes to the platform

richterb (approved):
🚀 🙏

PR #267, opened 2024-12-02
Chore: add accessor to formatters

etienneburdet:
## Summary

The goal for this PR is to add accessor functions to table fields, in order to extract the value from a more complex datasource. Overall it decouples what we pass as `value` to what we pass to `dsiplay`. It's needed for files on the platform, will be needed to display label on geocolumns and could be used for links etc.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-50639](https://app.shortcut.com/opendatasoft/story/feature/sc-50639/table-render-file-fields).

### Changes
The main change is to add an `accessor` functions, that transform rawValues into the value of the cell–which can now be different from it's display.
Other possibility: Tanstack table has `accessorKey` and `accessorFn` to quickly access another column. I though for our case, `accessorFn` would be general enough.

I also added an option to enable debug warning, because my console was exploding on the platform side 💥 

### Changelog
"We added accessor function to formatter (and thus table cells) so that it's possible to more complex object and extract value, e.g. `{href, tilte}`."

## Review checklist

- [ ] Description is complete


KevinFabre-ods (commented):
I struggle to understand the difference between `display` and `accessor` for some formats
To me, it's the same thing for Boolean, Short and Long text format. 

Can't we find another wording for this property? (like `transformer`?) and only defined it where it make sense.
- DateFormat
- NumberFormat
- URLFormat
- GeoFormat

Looking at the TanStack documentation, the `accessorfn` has the record item as 1st parameter. 
So It can be used to build something like this:

```js
item => `${item.first} ${item.lastName}`
```

In our case, the `accessor` only get the raw value as parameter.
As I said early maybe the wording is not quite right.

For complex formats such as DateFormat, NumberFormat, URLFormat, or GeoFormat, it makes sense to separate how the 'label' is constructed from how the raw value is used to render a date, number, URL link, or map. 
However, something still feels off to me. Perhaps the [comment](https://github.com/opendatasoft/ods-dataviz-sdk/pull/265#discussion_r1867364885) I made earlier regarding code reorganization, particularly around type checks and warning triggers, could help make this distinction clearer.

etienneburdet:
So I changed things a bit so that: 
* A column has an accessor for its value of form `fullRecord => value`
* Options can be either an object or an accessor of form `fullRecord => optionsObject`
* Formatters don't take accessors, but some options still are function of the rawValue.
* 
* I think it makes sense to have display mutating the value after everything (or else why using a formatter?). `sources` is very likely to be a function too, since rawValue will be a shape. In all cases it's always possible to fallback. to the accessor and make things that depend on the whole record.

I changed the geosshape story a bit, so that we make use of that to label the geofield with another column. I didn't implement "hidden" column option, but that's the spirit I think.

packages/visualizations/src/components/Format/URLFormat.svelte
```diff
@@ -1,28 +1,32 @@
...
+                text: display ? display(rawValue) : v,
                 href: v,
             };
         }
-        // eslint-disable-next-line no-console
-        console.warn(`ODS Dataviz SDK - Table: no url detected in ${v}. Formatting as string.`);
+
+        if (debugWarnings) {
+            warn(v, 'url');
+        }
```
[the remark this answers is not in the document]
etienneburdet: I don't think we could enable it component by component if it was a store. My problem was mainly to have hundreds of warnings (10 * 10 table is 100 cell already) or none at all when using it.

KevinFabre-ods (commented):
Hello, 

Could you try to update conditions where the value type is checked?
Like we agreed in https://github.com/opendatasoft/ods-dataviz-sdk/pull/265#discussion_r1867364885

etienneburdet:
Random naming though: `value/valueToString` instead of `rawValue/display` ?

KevinFabre-ods:
> Random naming though: `value/valueToString` instead of `rawValue/display` ?

Yes, that would be better.
Maybe `valueToLabel` `valueToText` is better than `valueToString` to clearly indicates that is will transform the label into something readable for a human

packages/visualizations/src/components/Table/types.ts
```diff
@@ -34,22 +35,32 @@ type BaseColumn = {
...
 
-type FormatPropsTypeMap = {
-    [DATA_FORMAT.boolean]: ColumnOptions<BooleanFormatProps>;
-    [DATA_FORMAT.date]: ColumnOptions<DateFormatProps>;
-    [DATA_FORMAT.geo]: ColumnOptions<GeoFormatProps>;
-    [DATA_FORMAT.shortText]: ColumnOptions<ShortTextFormatProps>;
-    [DATA_FORMAT.longText]: ColumnOptions<LongTextFormatProps>;
-    [DATA_FORMAT.number]: ColumnOptions<NumberFormatProps>;
-    [DATA_FORMAT.url]: ColumnOptions<URLFormatProps>;
+export type FormatPropsTypeMap = {
```
KevinFabre-ods, line 40: I got an error when I try to declare a column
`options.rawValue` shouldn't be required

image
```
Type '(r: Record<string, unknown>) => { display(v: string): string; }' is not assignable to type 'ValueOrAccessor<DateFormatProps> | undefined'.
  Type '(r: Record<string, unknown>) => { display(v: string): string; }' is not assignable to type '(r: Record<string, unknown>) => DateFormatProps'.
    Property 'rawValue' is missing in type '{ display(v: string): string; }' but required in type 'DateFormatProps'
 ```

**Review by KevinFabre-ods — COMMENTED** [review]
Great job to extract the format logic 👍 

I agree on the data flow from the table to the format components.

- `accessor` is for a table column to override the value we pass to a format  component ✅ 
- `column.options` is either a function (that take the record data) or an object ✅ 
   - `options.rawValue` is however required 🔴  For a column table, rawValue comes from `column.key` or `column.accessor`
- In format component, most props are primitive or object. If we have props that are function that take the result of the formatting as parameter ✅ 
      
I'm also in favor to rename `rawValue`

File: packages/visualizations/src/components/Format/index.ts [hunk 3]
```diff
@@ -1,30 +1,20 @@
...
-import LongTextFormat from './LongTextFormat.svelte';
 import NumberFormat from './NumberFormat.svelte';
-import ShortTextFormat from './ShortTextFormat.svelte';
+import TextFormat from './TextFormat.svelte';
 import URLFormat from './URLFormat.svelte';
-import { isValidRawValue } from './utils';
+import { isValidValue } from './utils';
 
 const Format = {
     boolean: BooleanFormat,
```
KevinFabre-ods, line 10: This Format constant doesn't seem to be used somewhere. 
Do we need it in our application ?

packages/visualizations/src/components/Table/types.ts
KevinFabre-ods: I still have the error. Can also see when running `npm run watch`
 > Type 'Record<string, unknown>' is not assignable to type 'Record'
 
 Shouldn't Column type be generic ?

packages/visualizations/src/components/Table/types.ts
```diff
@@ -34,22 +35,32 @@ type BaseColumn = {
...
+    [DATA_FORMAT.longText]: string;
+    [DATA_FORMAT.number]: number;
+    [DATA_FORMAT.url]: string;
 };
 
 export type ColumnOfType<F extends DataFormat> = BaseColumn & {
     dataFormat: F;
-    options?: FormatPropsTypeMap[F];
+    accessor?: (r: Record<string, unknown>) => ReturnTypeMap[F];
+    options?: ValueOrAccessor<FormatPropsTypeMap[F]>;
```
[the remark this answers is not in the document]
KevinFabre-ods: It's better but option is still defined as any
image

KevinFabre-ods (changes requested):
There are things to fixed with the debug data flow

Main issue is for me the typescript errors, as it stands, we'll get these errors too in our platform repository

packages/visualizations/src/components/Format/TextFormat.svelte
```diff
@@ -0,0 +1,23 @@
...
+        }
+        if (debugWarnings) {
+            warn(v, 'text');
+        }
+        return v;
+    };
+</script>
+
+<!-- Wrap value to style properly line clamp -->
+<span>{format(value)}</span>
```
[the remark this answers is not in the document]
etienneburdet: I think we have too if we want to be able to add text-ellipsis etc. with CSS override. Maybe we have to do that as well for the other? We just didn't run into the need yet.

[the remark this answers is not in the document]
KevinFabre-ods: IMO It depends on the "philosophy" we want to have here.

Do the format components only deal with the format value, or do they also set some display rules?

etienneburdet:
Ok so I fixed the options import by not using the path alias… they don't work with types between the two repos for now.

As for typing the accessor functions, I resorted to some `any`: doesn't yell at you if you don't input anything, but you can type the columns if you want. Combinations of `never ? … : ` or `R = unknown` just forced to add a generic all the way up to TableProps which was pretty alien.

packages/visualizations-react/stories/Table/options.ts
```diff
@@ -41,15 +55,15 @@ export const columns: Column[] = [
...
+            valueToLabel: (v: boolean) => (v ? 'αληθές' : 'ψευδές'),
         },
     },
     {
         title: 'Word count',
         key: 'wordCount',
         dataFormat: 'number',
         options: {
-            display: (v: string) => `${v} words`,
+            valueToLabel: (v: number) => `${v} words`,
```
KevinFabre-ods, line 66:
```suggestion
            valueToLabel: (v: string) => `${v} words`,
```
I have a TS error: `Type 'string' is not assignable to type 'number'`

packages/visualizations/src/components/Format/NumberFormat.svelte
```diff
@@ -4,24 +4,25 @@
...
 
-    $: format = (v: unknown) => {
-        if (!Number.isFinite(v)) {
-            warn(v, 'number');
-            return v;
+    $: format = (v: number) => {
+        if (Number.isFinite(v)) {
+            if (debugWarnings) {
+                warn(v, 'number');
+            }
```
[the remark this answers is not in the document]
KevinFabre-ods: Still exist

We should only warn when value is not a finite number

packages/visualizations/src/components/Table/Cell/utils.ts
```diff
@@ -1,8 +1,27 @@
...
+export const getOptions = <K extends DataFormat>(
+    column: ColumnOfType<K>,
+    record: Record<string, unknown>
+): Omit<FormatPropsTypeMap[K], 'value'> | null => {
+    const { options } = column;
+    if (options) {
+        return typeof options === 'function' ? options(record) : options;
+    }
+    return null;
+};
```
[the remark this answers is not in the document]
KevinFabre-ods: Could you please check that some code is missing to propagate the `debugWarnings` Table prop until the format components?

packages/visualizations/src/components/Table/TableCard.svelte
```diff
@@ -35,7 +39,14 @@
 
 <Card {title} {subtitle} {source} defaultStyle={!unstyled}>
     <div class="table-container">
-        <Table {loadingRowsNumber} {records} {columns} {description} {emptyStateLabel} {rowProps} />
+        <Table
+            columns={debugWarnings ? addWarnings(columns) : columns}
```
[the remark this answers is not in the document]
KevinFabre-ods: Currently, the only place where `addWarnings` is used will always get columns with `debugWarnings` set to false whether or not `tableOptions.debugWarnings` is either true or false

packages/visualizations/src/components/Format/TextFormat.svelte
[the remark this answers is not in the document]
KevinFabre-ods: Currently `TextFormat.Svelte` doesn't hold any style. 

The style definition is in `Cell.Svelte`.
What are our usage for this component outside of this package? 
If we use `TextFormat.Svelte` are we expecting some style attached to it?

packages/visualizations/src/components/Format/utils.ts
```diff
@@ -1,10 +1,19 @@
-export function isValidRawValue(rawValue: unknown): boolean {
-    return rawValue !== undefined && rawValue !== null;
+export function isValidValue(value: unknown): boolean {
+    return value !== undefined && value !== null;
```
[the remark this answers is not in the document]
KevinFabre-ods: ⏳

packages/visualizations/src/components/Table/types.ts
```diff
@@ -34,42 +34,65 @@ type BaseColumn = {
...
-export type ColumnOfType<F extends DataFormat> = BaseColumn & {
+/** Columns have to be typed with the record type if using an accessor.
+ * They can also be used without a record, nor accessor
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type ColumnOfType<F extends DataFormat, R = any> = BaseColumn & {
     dataFormat: F;
-    options?: FormatPropsTypeMap[F];
+    accessor?: (r: R) => ReturnTypeMap[F];
+    options?: ValueOrAccessor<Omit<FormatPropsTypeMap[F], 'value'>, R>;
```
KevinFabre-ods, line 66:
```suggestion
    accessor?: (r: R) => FormatPropsTypeMap[F]['value'];
    options?: ValueOrAccessor<Omit<FormatPropsTypeMap[F], 'value'>, R>;
```

With that `ReturnTypeMap` can be removed

KevinFabre-ods (commented):
I think we are done with all the new accessor logic. GG! 👏 

Could you please check my comments in my previous reviews, IMO we still need to resolve issues / suggestions

KevinFabre-ods (commented):
👋 

With your latest commit, I was able to resolve some of comments.

Could you give your thoughts about the comments that are still opened ?
- debugs warning are not thrown in the console https://github.com/opendatasoft/ods-dataviz-sdk/pull/267#discussion_r1882267437
- addition of the `debugWarnings` property can be simplified. there is one ternary condition too many https://github.com/opendatasoft/ods-dataviz-sdk/pull/267#discussion_r1905742467
- Unused variables https://github.com/opendatasoft/ods-dataviz-sdk/pull/267#discussion_r1882347130
- Should the `TextFormat` have a `span` tag ? https://github.com/opendatasoft/ods-dataviz-sdk/pull/267#discussion_r1882290266

KevinFabre-ods (approved):
GG WP! 👏

package.json
```diff
@@ -37,7 +37,7 @@
   },
   "devDependencies": {
     "lerna": "^8.1.2",
-    "nx": "^19.2.3"
+    "nx": "^20.2.1"
```
richterb, line 40: In general, better to avoid upgrades (especially a major version) in a feature PR 😄 But if you tested alright and Kevin and I didn't get any issue during review either, I guess we're good

packages/visualizations/src/components/Format/utils.ts
```diff
@@ -1,10 +1,19 @@
...
 }
 
 export function warn(value: unknown, format: string) {
-    // eslint-disable-next-line no-console
-    console.warn(`ODS Dataviz SDK - Table: ${value} is not a valid ${format}`);
+    switch (format) {
```
richterb, line 6: Usually I tend to add the "if (debugWarning)" in the debug function itself, so that future devs that use it don't have to remember to do the check themselves. Also it prevents having to pass `debugWarning` deep down. But I'm not sure that the store that holds the value is accessible here :/ My Svelte is rusty

richterb (approved):
The changes make sense! I added a small remark regarding the debug logger but it's minor and probably impossible to do 🙈

package.json
etienneburdet: It's for github actions, it crashes because nx doesn't find its arm64 or linux package… It's not the first time and I really don't know how to fix that. Maybe we should have a fixed version instead of carret 🤔

packages/visualizations/src/components/Format/utils.ts
etienneburdet: We have to subscribe manually, but `debugWarnings` store could be used in utils.ts indeed. (it changes in svelte 5 with .svelte.ts files).

package.json
richterb: Agreed. I once had a similar issue locally, and then not anymore a few days later, so that's a good probability 😬

wmai (approved):
I'm currently getting familiar with the `Table` component code, but at first glance, everything looks great 👏

I've only one small question regarding the `GeoFormat`, when `valueToLabel` is `undefined`, nothing is displayed. Is this intentional? Wouldn't it be preferable to default to displaying the coordinates as a simple string?

etienneburdet:
@wmai merging without a default for geo because as of now we pass a full maplibre style which really aweful, but also includes some variable names for sources and coordinates are children of it.

PR #270, opened 2024-12-26
feat: add onFeatureClick callback to map

etienneburdet:
## Summary

The goal for this PR is to add generic callback when clicking on a feature.

### Changes
It required adding both a callback and an array of layers, but I don't see how we could skip both those informations.

I contemplated mixing it with the popup layer somehow, but I think there are cases where want both Maplibre's Popup and a callback that does something else—although the main use case is alternative "popup".

### Changelog
Webgl maps now accept a callback when clicking featuers.

### DIFF
--- a/packages/visualizations/src/components/Map/WebGl/types.ts
+++ b/packages/visualizations/src/components/Map/WebGl/types.ts
@@ -4,6 +4,7 @@ import type {
     GeoJSONFeature,
     GestureOptions,
     LngLatLike,
+    MapGeoJSONFeature,
     RequestTransformFunction,
     StyleImageMetadata,
     StyleSpecification,
@@ -19,6 +20,11 @@ export type WebGlMapData = Partial<{
     layers: Layer[];
 }>;
 
+export type FeatureClickHandler = (f: MapGeoJSONFeature[]) => void;
+export type OnFeatureClick = {
+    callback: FeatureClickHandler;
+    layers: string[];
+};
 export interface WebGlMapOptions {
     /*
      * To render a basemap. Could be:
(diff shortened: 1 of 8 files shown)


packages/visualizations/src/components/Map/WebGl/types.ts
```diff
@@ -52,6 +58,8 @@ export interface WebGlMapOptions {
     preserveDrawingBuffer?: boolean;
     /** Images to load by the Map. keys are image ids  */
     images?: Images;
+    /** An addiational callback to be triggered when clicking a feature */
```
richterb, line 61: tiny typo

```suggestion
    /** An additional callback to be triggered when clicking a feature */
```

richterb:
Bonus during review: when testing it "for real", I thought it would be useful to have the cursor change when you hover something clickable. Turns out it's not trivial with Maplibre (according to https://stackoverflow.com/questions/55560489/mapbox-gl-on-mouse-hover-on-layers-change-cursor-pointer-style). But maybe we can have something in the component itself, so that when a layer has a callback, we're able to show a cursor when hovering an element of that layer, to spare re-users to have to handle it. I'm not sure whether `mouseenter` applies to the entire layer, or to anything that belongs to it, so maybe it's not even possible.

RafaelSzmarowski (approved):
LGTM 🚀

etienneburdet:
I added the cursor on layers that have either onFeature click or popup, it should do for now.

KevinFabre-ods:
> I added the cursor on layers that have either onFeature click or popup, it should do for now.

Hello, 

Why not update the `onMouseMove` function in `WebGl/WebGl.ts` for any features in any layers we get from the configuration ?

etienneburdet:
> > I added the cursor on layers that have either onFeature click or popup, it should do for now.
> 
> Hello,
> 
> Why not update the `onMouseMove` function in `WebGl/WebGl.ts` to set the cursor as pointer for any features in any layers we get from the configuration ?

That would be better indeed, pushed a fix.

richterb (approved):
LGTM 🚀 (minus the stuff you added for the beta obviously 😄)

PR #271, opened 2025-02-06
chore: scope sticky stores

etienneburdet:
## Summary

The goal for this PR is to scope the sticky stores so that the state isn't shared between all the instances of the components. 

We didn't see it much, because stickyness is reset in most cases. But in cases like navigating between pages with tables, the "isScrolled" state could be preserved. We don't have yet any place where multiple table with sticky columns, but this would have crashed as well.

### Changes
Made a factory and istanciate the store in the root component and then pass it around with a context to avoid props drilling and benifit from derived state.

## Review checklist

- [x] Description is complete

### DIFF
--- a/packages/visualizations/src/components/Table/Cell/Td.svelte
+++ b/packages/visualizations/src/components/Table/Cell/Td.svelte
@@ -1,16 +1,23 @@
 <script lang="ts">
-    import { isHorizontallyScrolled, stickyColumnsOffset } from '../store';
-    import type { Column } from '../types';
+    import { getContext } from 'svelte';
+    import type { StickyStores, Column } from '../types';
     import { getStickyClasses, getStickyOffset } from '../sticky';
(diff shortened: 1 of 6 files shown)


richterb (approved):
Good catch about the shared store 😳 
Maybe I am lucky, but I also wasn't able to reproduce the famous bug where the sticky columns are half-scrolled if it's not the first one (the bug that was raised during last review and was associated with a Svelte internal behavior if I remember correctly).

And the fix for the image makes sense, I suspected as much (the image having its definitive dimensions after the tooltip initially decided its position).

Testing was fine both in storybook and platform 👍

RafaelSzmarowski (approved):
LGTM ! Same as @richterb I never noticed the shared sticky state bug but I don't think we have many places it could happen. Anyway the fix seems logical and as far as I have tested it, everything seems to work perfectly fine both for sticky columns and tooltips 🥇  Good job !

PR #273, opened 2025-02-12
fix: add an empty column in loading cells too

etienneburdet:
## Summary

The goal for this PR is to fix a case where loading cell would be misaligned with actual columns when a button column was present.

The cause is that we insert an extra column if `rowProps.onclick` is defined, but didin't do it on the loading part.

### Changes
Add an empty cell with a min-width of 28px (button width) so that column count and alignment is preserved.

I added it to the loading story too to have a snapshot.

## Review checklist

- [x] Description is complete


RafaelSzmarowski (approved):
Good to go 🚀

PR #274, opened 2025-02-18
chore: upgrade react

etienneburdet:
## Summary

The goal for this PR is to upgrade React to the platform version and provide better packaging so that we can actually require component in the tests when that makes sense (that is KPI, Formatters, Table).

The current problem is, umd takes priority over all other in jest builds  and that the umd does package react. So even with the same version, we end up with two react in jest…
`type: module` would be fine if it wasn't for `jest.requireActual(…)` which uses `require(…)` (no shit).

So the solution is to export CJS and ESM, but not UMD and use exports: import/require in package.json. Seems to be the most common practice.

## To be tested
Nx seems to be randomly failing on CI with code 135, I don't really know why. Maybe just fork this branch and try to run CI on separate PR to see if it's to severe or not. Else we might just turn it off, we barely use it and it's been buggy all the time on CI.


packages/visualizations-react/package.json
```diff
@@ -54,7 +55,9 @@
     },
     "dependencies": {
         "@opendatasoft/visualizations": "0.27.4",
-        "use-callback-ref": "^1.2.4"
+        "use-callback-ref": "^1.2.4",
+        "react": "^17.0.2",
+        "react-dom": "^17.0.2"
```
richterb, line 60: I don't think we want to package React in the lib itself 😬 Why did it move from devDependencies?

etienneburdet: It's just a try to see if it fixes test behavior in the platform (I tried other changes in the platform at the same time) 😅

packages/visualizations-react/package.json
```diff
@@ -105,16 +111,16 @@
...
+        "react-dom": "^17.0.2",
         "rollup": "^4.18.0",
         "rollup-plugin-postcss": "^4.0.2",
         "size-limit": "^8.1.0",
         "storybook": "^7.6.17",
         "storybook-addon-rtl": "^1.0.0",
         "tslib": "^2.1.0",
-        "typescript": "^5.5.3"
+        "typescript": "^5.5.3",
+        "vite": "^5.4.14"
```
richterb, line 123: 👀 Leftover from a test ? 😄

richterb:
In my memory, the main reason for UMD was so that we could use our packages in Observable. I believe they updated a bit, and support ES6 native imports now (https://observablehq.com/@observablehq/require#cell-106). Can you test a bit and see if this works? (via the spypack CDN method I guess since you published the beta on npm)

packages/visualizations-react/package.json
etienneburdet: No, vite is used by storybook now and I don't know why in the previous iteration it found its own vite but I had to install explicitely this time 🤔

etienneburdet:
> In my memory, the main reason for UMD was so that we could use our packages in Observable. I believe they updated a bit, and support ES6 native imports now (https://observablehq.com/@observablehq/require#cell-106). Can you test a bit and see if this works? (via the spypack CDN method I guess since you published the beta on npm)

I was about to test that but it's only the react package, not the core one, I changed. I don't think there's any practical way to use it in Observable—loading React there might lead to very strange things. Realistically I think the React package targets bundler.

packages/visualizations-react/rollup.config.mjs
```diff
@@ -121,6 +149,6 @@ const umd = defineConfig({
 });
 
 // Just compile the ESM version during development
-const bundles = production ? [esm, umd] : [esm];
+const bundles = production ? [esm, cjs] : [esm];
```
richterb, line 152: We can remove the `umd` config entirely from this file now I guess?

packages/visualizations-react/package.json
```diff
@@ -14,14 +14,20 @@
...
-    "main": "dist/index.umd.js",
+    "main": "dist/index.es.js",
     "module": "dist/index.es.js",
-    "unpkg": "dist/index.umd.js",
     "typings": "dist/src/index.d.ts",
     "files": [
         "dist",
         "src"
     ],
+    "exports": {
```
richterb, line 24: I get this error in platform: 
> Module not found: Error: Package path ./dist/index.css is not exported from package /usr/src/app/ods/react/node_modules/@opendatasoft/visualizations-react (see exports field in /usr/src/app/ods/react/node_modules/@opendatasoft/visualizations-react/package.json)

Adding `"./dist/": "./dist/"` seems to fix the warning indeed

richterb (commented):
A question regarding a warning I had, but otherwise clear and LGTM!

PR #275, opened 2025-02-28
[API-CLIENT][FEAT] Add metadata template field type and ODSQL lower function

KevinFabre-ods:
## Summary

The goal for this PR is to add utils to the api client package

### Changes

- Package has type and constant for metadata template field types
- Add a new ODSQL function lower

## Open discussion

About the lower function.
I think we'll mostly use the combination `lower(field(' my field'))` when building queries
  
Do you see any reasons to not run the `field()` function inside `lower()` ?

We can't call `lower` inside `field` because it is not supported in every clause.
  
Other possibility would be to have optional arguments when calling `orderBy` or  `groupBy` to include the lower function but that requires to rewrite mostly all the logic of the Query class.

### DIFF
--- a/packages/api-client/src/odsql/index.ts
+++ b/packages/api-client/src/odsql/index.ts
@@ -168,6 +168,17 @@ export const string = (value: string) => JSON.stringify(value);
 
 export const dateTime = (date: Date) => `date'${date.toISOString()}'`;
 
+/**
+ * ODSQL function to normalize field name.
+ *
+ * - Can be used for `order_by` and `group_by` clauses.
+ * - Example usage: `select=count(*) & group_by=lower(artist)`.
+ *
+ * @param {string} fieldName - The field name to normalize.
+ * @returns {string}  The lower ODSQL function.
+ */
+export const lower = (fieldName: string) => `lower(${fieldName})`;
+
 function inlineMonthOrDay(value?: number) {
     if (value === undefined) return '';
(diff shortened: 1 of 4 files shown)


packages/api-client/src/odsql/index.ts
```diff
@@ -168,6 +168,17 @@ export const string = (value: string) => JSON.stringify(value);
...
+/**
+ * ODSQL function to normalize field name.
+ *
+ * - Can be used for `order_by` and `group_by` clauses.
+ * - Example usage: `select=count(*) & group_by=lower(artist)`.
+ *
+ * @param {string} fieldName - The field name to normalize.
+ * @returns {string}  The lower ODSQL function.
+ */
+export const lower = (fieldName: string) => `lower(${fieldName})`;
```
KevinFabre-ods, line 180: We could include the result of the `field()` function
```suggestion
export const lower = (fieldName: string) => field(`lower(${fieldName})`);
```

KevinFabre-ods:
@richterb do you anything to add, or we are good to merge ?

richterb:
> @richterb do you anything to add, or we are good to merge ?

I wanted to spend a bit of time to find examples but I agree with William and I think we have risks/examples of having double layers of escaping quotes if we were to do that (I had some issues with the Query data tab similar to that). So the PR as is looks good to me!

PR #276, opened 2025-04-24
[FEAT][CHART] add TimeDisplayFormats for chart options

cybervinvin:
## Summary

The goal for this PR is to improve the `TimeCartesianAxisConfiguration` type to add `TimeDisplayFormats`.
By exposing `TimeDisplayFormats`, we allow the modification of the display format of the labels on the graph axis.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-55009](https://app.shortcut.com/opendatasoft/story/55009/as-a-user-i-want-to-display-my-data-results-by-weeks-time-scale).

### Changes

Improve the type `TimeCartesianAxisConfiguration`.

### Changelog

In the Chart component: It's now possible to update the display for the x axis when the type is "time".

## Review checklist

- [X] Description is complete

### DIFF
--- a/packages/visualizations/src/components/Chart/scales.ts
+++ b/packages/visualizations/src/components/Chart/scales.ts
@@ -61,7 +61,7 @@ const DATE_TOOLTIP_FORMATS = {
     minute: DateTime.DATETIME_MED,
     hour: DateTime.DATETIME_MED,
     day: { day: 'numeric', month: 'long' },
-    week: 'DD',
+    week: "'W'WW yyyy",
     month: { month: 'long', year: 'numeric' },
     quarter: "'Q'q - yyyy",
     year: { year: 'numeric' },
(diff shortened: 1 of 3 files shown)


packages/visualizations/src/components/Chart/scales.ts
```diff
@@ -93,6 +93,10 @@ export default function buildScales(options: ChartOptions): ChartJsChartOptions[
                       time: {
                           unit: options?.axis?.x?.timeUnit,
                           tooltipFormat: getDateTooltipFormat(options?.axis?.x?.timeUnit),
+                          isoWeekday: true,
```
richterb, line 96: I wondered if Arabic customers would like to have weeks starting on Sunday (because their work weeks start on Sunday), but in our AngularJS code, we never implemented such a thing, so I think we can start like this indeed.

richterb (approved):
Looks good to me!

Can you add a story dedicated to the "Week" case in `ChartTimeScale.stories.tsx` like we did for the other scales?

wmai (approved):

packages/visualizations/src/components/Chart/scales.ts
cybervinvin: I think so too but I can also delete it :neutral_face:

cybervinvin:
@richterb : added story ✅

packages/visualizations/src/components/Chart/scales.ts
richterb: No it's fine, the code you added makes the SDK behave exactly like the "old" charts and fits 99% of our customers 🙂 It was a remark in case some Arabic customer points it out in the future. If it happens, we'll probably make it configurable from the outside.

PR #279, opened 2025-06-12
[FIX] RTL - minors improvements

KevinFabre-ods:
## Summary

The goal for this PR is to fix style that weren't compatible with RTL

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-56648](https://app.shortcut.com/opendatasoft/story/56648).

### Changes
- Legend: update spacing between the symbol and the label (use flex `gap` instead of `margin-left`)
- Table: update spacing for the zoom icon button (use margin-inline-start/end instead of margin-left/right)
- Table: update spacing value for the icon in the page size select


KevinFabre-ods:
> This story appears now as broken: https://616594e84c1692003af430f3-qoowotnoat.chromatic.com/?path=/story/table-pagination--page-size
> 
> Capture d’écran 2025-07-07 à 16 08 47
> 
> Do you also see this ?

@RafaelSzmarowski It should be fixed now ;)

PR #280, opened 2025-06-30
feat(api-client): Support timezone parameter

richterb:
## Summary

The goal for this PR is to add support to the `timezone` parameter.

### Changes

Just add a `.timezone(<timezone>)` parameter, similar to the others like `lang` etc.

#### Breaking Changes

None

### Changelog

We added a new `timezone()` function on queries, to support the `timezone` parameter in API endpoints.

## Review checklist

- [x] Description is complete

### DIFF
--- a/packages/api-client/src/odsql/index.ts
+++ b/packages/api-client/src/odsql/index.ts
@@ -130,6 +130,10 @@ export class Query {
     lang(lang: string): Query {
         return this.set('lang', lang);
     }
+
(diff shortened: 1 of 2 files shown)


richterb:
The doc points to https://github.com/unicode-org/cldr/blob/main/common/bcp47/timezone.xml (it's in the doc of the `timezone` parameter of one of the endpoints, the search doesn't seem to work), I've only ever seen values in the `alias` attributes of this file

PR #292, opened 2026-04-01
fix(Table): Missing labels for accessibility

richterb:
## Summary

The goal for this PR is to add two missing labels, invisible in the page, but useful for screen readers and other accessibility user agents.

(Internal for Opendatasoft only) Associated Shortcut ticket: 
- [sc-60636](https://app.shortcut.com/huwise/story/60636)
- [sc-60637](https://app.shortcut.com/huwise/story/60637)

### Changes
- Added a label to the `th` that is added for the "extra button column" (used for "Open record" in the platform)
- Added a label to the `select` used to select the number of records per page in the pagination control
- Exposed the labels as properties so that we can inject translated terms from the platform

## To be tested
This can be tested in Storybook, using the browser inspector

## Review checklist

- [ ] Description is complete

### DIFF
--- a/packages/visualizations/src/components/Pagination/PageSize.svelte
+++ b/packages/visualizations/src/components/Pagination/PageSize.svelte
@@ -5,6 +5,7 @@
     export let options: PageSizeOption[];
     export let onChange: (size: number) => void;
     export let selected: number;
+    export let selectLabel = 'Number of records per page';
 
     let value: number;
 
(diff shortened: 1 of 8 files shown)


cybervinvin (approved):
LGTM 👍🏼 
I left one comment about a label key naming that could be confusing.
I think that a test checking that the `aria-label` is correctly rendered with and without the label prop provided is necessary to prevent regressions.

PR #294, opened 2026-04-16
feat: support action items and href onClick

wmai:
## Summary

`LinksMenu` now supports both navigational links and client-side actions in the same menu, with an optional click hook and download filename hint on links.

Internal context, [sc-59779](https://app.shortcut.com/opendatasoft/story/59779).

## Public API

`Link` is a discriminated union, `LinkHref | LinkAction`.

`LinkHref` renders `<a target="_blank" rel="noopener noreferrer">`:
- `href`, `label`, `icon?`
- `onClick?`, called before navigation (analytics on downloads)
- `download?`, hint for the browser's `download` attribute (same-origin only)

`LinkAction` renders `<button type="button">`:
- `onClick`, `label`, `icon?`

Helpers exported, `isLinkHref(link)` type guard, `LINKS_MENU_CLASS` constant (CSS root, lets consumers exclude the menu from DOM screenshots).

## Changelog

Feature, `LinksMenu` accepts action items alongside href links, with optional `onClick` and `download` on href links.

## Test plan

- One href link with an `onClick` spy, click it, expect spy called and menu closed
- One action item, click it, expect callback called and menu closed
- One href link with a `download` string, click it, browser uses the hint as filename (same-origin URL)

### DIFF
--- a/packages/visualizations/src/components/utils/LinksMenu.svelte
+++ b/packages/visualizations/src/components/utils/LinksMenu.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
     import { onDestroy, tick } from 'svelte';
     import type { LinksMenuProps } from 'types';
+    import { isLinkHref } from 'types';
 
     // Ensure exported type matches declared props
     type $$Props = LinksMenuProps;
@@ -10,7 +11,7 @@
 
     let menuElement: HTMLDivElement;
     let buttonElement: HTMLButtonElement;
-    let menuItemElements: HTMLAnchorElement[] = [];
+    let menuItemElements: (HTMLElement | undefined)[] = [];
     let isOpen = false;
     let focusedIndex = -1;
 
@@ -38,6 +39,11 @@
         } else {
             openMenu();
         }
+    }
+
+    function handleMenuItemClick(link: $$Props['links'][number]) {
+        link.onClick?.();
+        closeMenu();
     }
 
     function focusItem(index: number) {
@@ -173,23 +179,41 @@
             on:keydown={handleMenuKeydown}
         >
             {#each links as link, index}
-                <a
-                    bind:this={menuItemElements[index]}
-                    href={link.href}
-                    target="_blank"
-                    rel="noopener noreferrer"
-                    class="dropdown-item"
-                    role="menuitem"
-                    tabindex={focusedIndex === index ? 0 : -1}
-                    on:click={() => closeMenu()}
-                >
-                    {#if link.icon}
-                        <span class="dropdown-item-icon" aria-hidden="true">
-                            {@html link.icon}
-                        </span>
-                    {/if}
-                    <span class="dropdown-item-label">{link.label}</span>
-                </a>
+                {#if isLinkHref(link)}
+                    <a
+                        bind:this={menuItemElements[index]}
+                        href={link.href}
+                        target="_blank"
+                        rel="noopener noreferrer"
+                        class="dropdown-item"
+                        role="menuitem"
+                        tabindex={focusedIndex === index ? 0 : -1}
+                        on:click={() => handleMenuItemClick(link)}
+                    >
+                        {#if link.icon}
+                            <span class="dropdown-item-icon" aria-hidden="true">
+                                {@html link.icon}
+                            </span>
+                        {/if}
+                        <span class="dropdown-item-label">{link.label}</span>
+                    </a>
+                {:else}
+                    <button
+                        bind:this={menuItemElements[index]}
+                        type="button"
+                        class="dropdown-item dropdown-item--action"
+                        role="menuitem"
+                        tabindex={focusedIndex === index ? 0 : -1}
+                        on:click={() => handleMenuItemClick(link)}
+                    >
+                        {#if link.icon}
+                            <span class="dropdown-item-icon" aria-hidden="true">
+                                {@html link.icon}
+                            </span>
+                        {/if}
+                        <span class="dropdown-item-label">{link.label}</span>
+                    </button>
+                {/if}
             {/each}
         </div>
     {/if}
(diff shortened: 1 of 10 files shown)


packages/visualizations/src/components/utils/LinksMenu.svelte
```diff
@@ -289,4 +314,17 @@
...
+        width: 100%;
+        margin: 0;
+        border: none;
+        background: transparent;
+        font: inherit;
+        text-align: left;
+        cursor: pointer;
+        appearance: none;
+        -webkit-appearance: none;
+    }
```
RafaelSzmarowski, line 334: One difference noticed, this item has no font-size 14px contrary to the .dropdown-item above which ends up in a difference. I just want to confirm that this is intended ? Probably not and due to `font: inherit` on the button that overrides `.dropdown-item’s font-size: 14px`. Probably we should use more granular overrides: `font-family: inherit; font-weight: inherit; line-height: inherit;` and not the global `font: inherit`

Image

packages/visualizations/src/types.ts
```diff
@@ -6,18 +6,44 @@ export type DataFrame = Record<string, any>[];
...
+ * Menu item in LinksMenu: either a navigational link or an action.
+ */
+export type Link = LinkHref | LinkAction;
+
+export function isLinkHref(link: Link): link is LinkHref {
+    return 'href' in link;
+}
+
+/** Root CSS class of the LinksMenu wrapper. Consumers can use this to filter the menu from DOM clones (e.g. image export). */
+export const LINKS_MENU_CLASS = 'links-menu';
```
RafaelSzmarowski, line 50: As we have this new constant and exported for targetting, we should reuse this constant in LinksMenu.svelte instead of the hardcoded string to avoid future possible divergence

packages/visualizations/src/components/utils/LinksMenu.svelte
```diff
@@ -40,6 +41,11 @@
         }
     }
 
+    function handleMenuItemClick(link: $$Props['links'][number]) {
+        link.onClick?.();
+        closeMenu();
+    }
```
RafaelSzmarowski, line 50: A bit optional but I just wonder if we want the onClick to block the close menu if something goes wrong. Maybe this would prevent: 
```suggestion
    function handleMenuItemClick(link: ...) {
    try {
        link.onClick?.();
    } finally {
        closeMenu();
    }
}
```

packages/visualizations/src/types.ts
```diff
@@ -6,18 +6,44 @@ export type DataFrame = Record<string, any>[];
...
  */
-export interface Link {
-    /** The URL to navigate to when the link is clicked */
+export interface LinkHref {
     href: string;
-    /** The label displayed for this link */
     label: string;
     /** Optional SVG string to display as an icon next to the label */
     icon?: string;
+    /** Optional click callback invoked before opening the href. */
```
RafaelSzmarowski: Maybe worth just a small precision: this is side effect only it does not replace, prevent or stop navigation which happens after.

packages/visualizations/src/components/utils/LinksMenu.svelte
```diff
@@ -1,6 +1,7 @@
 <script lang="ts">
```
RafaelSzmarowski, line 1: Maybe worth adding some basic tests for this component to cover main cases

packages/visualizations/src/components/utils/LinksMenu.svelte
RafaelSzmarowski, line 180: Outside this PR but the aria-label is a hardcoded english "links" and also now it can be also an action item. Probably we could improve this but as this is a bit outside it's optional

RafaelSzmarowski (changes requested):
Looks good and works well ! I just have one real change request which is the mismatch of font sizes, the rest is more debatable and clearly optional

packages/visualizations/src/components/utils/LinksMenu.svelte
wmai: Fixed, and you nailed the cause : [here](https://github.com/opendatasoft/ods-dataviz-sdk/blob/b9ffe2803dbd0467b4510846c3fda8511fb24f14/packages/visualizations/src/components/utils/LinksMenu.svelte#L322-L334)

packages/visualizations/src/types.ts
wmai: Done [here](https://github.com/opendatasoft/ods-dataviz-sdk/blob/b9ffe2803dbd0467b4510846c3fda8511fb24f14/packages/visualizations/src/components/utils/LinksMenu.svelte#L149).

packages/visualizations/src/components/utils/LinksMenu.svelte
wmai: Took your suggestion as-is [here](https://github.com/opendatasoft/ods-dataviz-sdk/blob/b9ffe2803dbd0467b4510846c3fda8511fb24f14/packages/visualizations/src/components/utils/LinksMenu.svelte#L44-L50)

packages/visualizations/src/types.ts
wmai: Done [here](https://github.com/opendatasoft/ods-dataviz-sdk/blob/b9ffe2803dbd0467b4510846c3fda8511fb24f14/packages/visualizations/src/types.ts#L16)

packages/visualizations/src/components/utils/LinksMenu.svelte
wmai: Added [here](https://github.com/opendatasoft/ods-dataviz-sdk/commit/4588dbef6fbd77f61ccbb791fc76f740fdf4d6c3)

RafaelSzmarowski (approved):
LGTM ! Thanks for the last changes 🚀

PR #295, opened 2026-05-19
fix(Table): Screen-reader visible header for actions columm

richterb:
## Summary

The goal for this PR is to fix accessibility labels for the "actions" column of the Table component.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-61551](https://app.shortcut.com/opendatasoft/story/61551).

### Changes
As described in the story, using aria-label is not a recommended practice, so I got back to the usual `sr-only`. Alternatively we could even display the header itself for real, but it's a design discussion we can have later.

It felt like a good idea to make an utility class shared across the code, so I made one. However, in order to have a clean import (`styles/accessibility`), I had to make sure CSS imports were resolved against `src` both in rollup (for build), and in svelte config (for svelte-check), so it pulled a bit more changes. If it feels a bit too much, we could always have imports like `../../../styles/accessibility` which is not that much of a problem, considering a mistake would be caught immediately.

## To be tested
The `th` for the actions column now has a real label hidden for browser but not for screen readers, instead of an aria-label.

## Review checklist

- [ ] Description is complete

### DIFF
--- a/packages/visualizations/rollup.config.mjs
+++ b/packages/visualizations/rollup.config.mjs
@@ -2,7 +2,7 @@ import path from 'path';
 import { fileURLToPath } from 'url';
 import svelte from 'rollup-plugin-svelte';
 import alias from '@rollup/plugin-alias';
-import autoPreprocess from 'svelte-preprocess';
+import svelteConfig from './svelte.config.mjs';
 import postcss from 'rollup-plugin-postcss';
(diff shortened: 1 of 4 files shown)


RafaelSzmarowski (approved):
Works as expected, I tested with VoiceOver and could access the `sr-only` span 🚀

PR #296, opened 2026-05-19
feat(Table): add sticky header option for Table

RafaelSzmarowski:
## Summary

The goal for this PR is to add a sticky header option for the Table component.

The ticket mentions sticky rows but the real use case is only for the header. So I decided to keep it simple and only add the property for the header. Making a generic sticky rows was quite difficult and seemed probably not worth the effort (how to integrate pagination and sticky rows, how to handle properly table height changes, we needed to add to the header th and to the body the sticky property)

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-61533](https://app.shortcut.com/huwise/story/61533/explorer-freeze-rows-support-in-ods-dataviz-sdk-table-component).


RafaelSzmarowski:
@wmai @cybervinvin I'll switch this PR to draft until we have a clear design vision and see if this is sufficient

RafaelSzmarowski:
Maybe need a rebase but after discussion we'll go with a sticky header only. So the PR is ready again

RafaelSzmarowski:
@wmai and @cybervinvin you can do the review, after I'll rebase

packages/visualizations/src/components/Table/Table.svelte
```diff
@@ -15,25 +15,37 @@
...
+        isVerticallyScrolled,
     });
 
     function handleScroll() {
         $isHorizontallyScrolled =
             document.dir === 'rtl' ? scrollBox?.scrollLeft < 0 : scrollBox?.scrollLeft > 0;
+        $isVerticallyScrolled = (scrollBox?.scrollTop ?? 0) > 0;
+    }
+
+    $: if (scrollBox) {
```
cybervinvin, line 47: This new block and `handleScroll()` seem to update `$isVerticallyScrolled` in two different ways. Is this block needed on top of `handleScroll()`, or is it a leftover? If it's intentional, a short comment on why would help future readers.

packages/visualizations/src/components/Table/Headers/Headers.svelte
```diff
@@ -5,15 +5,16 @@
     export let columns: Column[];
     export let extraButtonColumn = false;
     export let extraButtonColumnLabel: string | undefined;
+    export let stickyHeader = false;
```
cybervinvin, line 8: `stickyHeader` is a single table-level boolean, but it's threaded as a prop through `Table` -> `Headers` -> `Th`, while `isVerticallyScrolled`, added right next to it in this same PR, goes through the existing `sticky-stores` context instead. That leaves `stickyHeader` with 5 separate default declarations across the codebase (`Table.svelte`, `TableCard.svelte`, `Headers.svelte`, `Th.svelte`, `sticky.ts`). Any reason not to put it in context alongside `isHorizontallyScrolled` and `isVerticallyScrolled`? That would remove the prop threading in `Headers.svelte` and the `export let` in `Th.svelte`.

packages/visualizations/src/components/Table/sticky.ts
```diff
@@ -7,20 +7,31 @@ export const getStickyClasses = ({
     column,
```
cybervinvin, line 7: `getStickyClasses` is a pure function and this PR roughly doubles its branching (column sticky state x stickyHeader x verticallyScrolled x lastStickyColumn). It has no unit tests today, consistent with the rest of `sticky.ts`/`store.ts`, so not a regression from this PR specifically. Worth adding coverage now while the branching is fresh, or was that deliberately left to the Storybook stories?

cybervinvin:
In Storybook (StickyHeaderAndColumns story): scroll the table vertically first, then horizontally. The header's bottom shadow seems to shift/wave on sticky column headers while scrolling horizontally, instead of staying a stable line.

3_debut_scroll_horizontal

Hypothesis: every header <th> gets its own independent box-shadow here once isVerticallyScrolled is true, including both sticky columns (fixed on screen via position: sticky; left: ...) and non-sticky columns (which visually scroll underneath them). Since the sticky columns' shadow stays fixed while the non-sticky columns' shadow moves with horizontal scroll, their overlap changes continuously as you scroll, which reads as a wave. If that's the cause, the fix is probably not local to this rule. It likely needs the header shadow applied as one layer above the whole header row rather than per cell, so it isn't split across cells scrolling at different visual speeds. Can you confirm?

cybervinvin (changes requested):
In Storybook (StickyHeaderAndColumns story): scroll the table vertically first, then horizontally. The header's bottom shadow seems to shift/wave on sticky column headers while scrolling horizontally, instead of staying a stable line.

3_debut_scroll_horizontal

Hypothesis: every header <th> gets its own independent box-shadow here once isVerticallyScrolled is true, including both sticky columns (fixed on screen via position: sticky; left: ...) and non-sticky columns (which visually scroll underneath them). Since the sticky columns' shadow stays fixed while the non-sticky columns' shadow moves with horizontal scroll, their overlap changes continuously as you scroll, which reads as a wave. If that's the cause, the fix is probably not local to this rule. It likely needs the header shadow applied as one layer above the whole header row rather than per cell, so it isn't split across cells scrolling at different visual speeds. Can you confirm?

packages/visualizations/src/components/Table/Table.svelte
RafaelSzmarowski: Good catch, it was a leftover and removed. Now only handleScroll updates $isVerticallyScrolled now.

packages/visualizations/src/components/Table/Headers/Headers.svelte
RafaelSzmarowski: Agreed and real improvement indded. `stickyHeader` now lives in the sticky-stores context next to the scroll flags, so Headers andTh no longer take it as a prop.

packages/visualizations/src/components/Table/sticky.ts
RafaelSzmarowski: I dropped the extra branching from getStickyClasses (header stickiness andshadow moved to the thead). sticky.ts is back to its previous shape, so no new coverage for this PR.

RafaelSzmarowski:
> In Storybook (StickyHeaderAndColumns story): scroll the table vertically first, then horizontally. The header's bottom shadow seems to shift/wave on sticky column headers while scrolling horizontally, instead of staying a stable line.
> 
> 3_debut_scroll_horizontal
> Hypothesis: every header gets its own independent box-shadow here once isVerticallyScrolled is true, including both sticky columns (fixed on screen via position: sticky; left: ...) and non-sticky columns (which visually scroll underneath them). Since the sticky columns' shadow stays fixed while the non-sticky columns' shadow moves with horizontal scroll, their overlap changes continuously as you scroll, which reads as a wave. If that's the cause, the fix is probably not local to this rule. It likely needs the header shadow applied as one layer above the whole header row rather than per cell, so it isn't split across cells scrolling at different visual speeds. Can you confirm?

Thanks and good hypothesis. Normally fixed by applying sticky + shadow on the thead as one layer instead of per cell.

RafaelSzmarowski:
@cybervinvin thanks for the review, I've made all the requested changes, you can have a second look !

RafaelSzmarowski:
@wmai and @cybervinvin I added a new `maxHeight` prop to Table options in my last commit: https://github.com/opendatasoft/ods-dataviz-sdk/pull/296/commits/853b1ee0e404c57e07fe7b84929e4db65c7caf4f

Because I noticed that the sticky header option by itself was not enough. The `thead` sticks against `.scrollbox`, which declares `overflow-x: auto`. It just never has a bounded height, so nothing ever scrolls inside it and the header never pins. A consumer would have to go and target the global CSS to fix a max height to make scrollbox scrollable. This way it felt cleaner, so to make a Table scrollable and use the sticky header we just need to pass `stickyHeader: true` and `maxHeight` to the Table options.

packages/visualizations-react/stories/Table/StickyHeader.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+                    />
+                </div>
+            </>
+        );
+    },
+};
+
+export const StickyHeaderWithPagination: StoryObj<typeof Table> = {
+    args: {
+        current: 1,
```
wmai, line 111: `tsc` is unhappy here (`current` is not on `TableProps`, and the `render` type does not match). They are the only two errors in the package. Nothing in CI runs `tsc`, and the `check: true` in `.storybook/main.js` needs `vite-plugin-checker`, which we do not have, so the PR stays green anyway. The pagination stories type against the template instead of against `Table`, that would clear it.

packages/visualizations/src/components/Table/Table.svelte
```diff
@@ -15,25 +15,41 @@
...
 
+    $: $isHeaderSticky = stickyHeader;
+    /* Set inline rather than in the scoped style block: `unstyled` drops the
+       .ods-dataviz--default rules, and the scrollport must be bounded either way. */
+    $: scrollboxStyle = maxHeight ? `max-height: ${maxHeight}; overflow-y: auto;` : undefined;
+
     function handleScroll() {
         $isHorizontallyScrolled =
             document.dir === 'rtl' ? scrollBox?.scrollLeft < 0 : scrollBox?.scrollLeft > 0;
+        $isVerticallyScrolled = scrollBox?.scrollTop > 0;
```
cybervinvin, line 58: The reset a bit further down only runs when `columns` changes, not when `records` changes. On pagination, `scrollBox.scrollTop` (and `scrollLeft`) carry over from the previous page, so the new page can render already scrolled from the very first frame, even though the user never touched it there. The `StickyHeaderWithPagination` story reproduces it: scroll down on page 1, then change page. Worth resetting `scrollTop`/`scrollLeft` here too when `records` changes?

packages/visualizations-react/stories/Table/StickyHeader.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+        const [isStickyHeaderEnabled, setIsStickyHeaderEnabled] = useState(
+            Boolean(templateOptions.stickyHeader)
+        );
+
+        return (
+            <>
+                <div style={{ marginBottom: '20px' }}>
+                    <h3>Sticky header (`options.stickyHeader`)</h3>
+                    <p style={{ maxWidth: '42rem', marginBottom: '0.75rem' }}>
+                        Scroll inside the table — the header row stays pinned at the top.
```
cybervinvin: `stickyHeader` has no effect unless `maxHeight` is also set (already documented at the type level in `types.ts`). Worth a one-line mention here too, since Storybook is probably where most integrators will look first, e.g. "Requires `maxHeight` to be set too, otherwise the header has no visible effect."

wmai (changes requested):
Looks good 👏 Works like charm.

One small TypeScript issue in the story file, nothing on the component itself.

RafaelSzmarowski:
@wmai and @cybervinvin Thanks again for the review, I did the request changes ! You can have another look 😃

cybervinvin (approved):
@RafaelSzmarowski Thanks for addressing everything, went through a second round with fresh eyes, all good on my end. Nice work, approving.

wmai (approved):
🚀

RafaelSzmarowski:
@wmai and @cybervinvin I just added a new prop fillHeight to align with how we offer this option in our charts. It will be used directly in our use case in the platform. Just setting the parent container to 100% and passing the fillHeight prop to the Table. I hesitated but still kept the `maxHeight` prop as well, I thought it kind of makes sense with the stickyHeader, if we decide not leaving this option, I can remove it.

packages/visualizations/src/components/Table/Table.svelte
```diff
@@ -25,19 +28,43 @@
...
         lastStickyColumn,
         isHorizontallyScrolled,
+        isVerticallyScrolled,
+        stickyHeader: isHeaderSticky,
     });
 
+    $: $isHeaderSticky = stickyHeader;
+    /* Length caps stay inline: `unstyled` drops the .ods-dataviz--default rules,
+       and the scrollport must be bounded either way. */
+    $: scrollboxStyle = maxHeight ? `max-height: ${maxHeight}; overflow-y: auto;` : undefined;
```
wmai: If both options are set, the card fills the container but the rows stop at `maxHeight`, so you get empty space below. Maybe skip `maxHeight` when `fillHeight` is on?

Image

```suggestion
    $: scrollboxStyle =
        maxHeight && !fillHeight ? `max-height: ${maxHeight}; overflow-y: auto;` : undefined;
```

After the fix :

Image

RafaelSzmarowski: Thanks and good catch ! I made your suggestion here: https://github.com/opendatasoft/ods-dataviz-sdk/pull/296/commits/c96d2822c9890824dbc8350622ad0a6675314410

wmai (approved):
LGTM 🤌

cybervinvin:
The `fillHeight` addition looks solid, no blockers on my end. Good work.

cybervinvin (approved):

PR #298, opened 2026-06-15
[sc-61255] feat: Table cursor pagination

wmai:
## Summary

The `Table` pager currently needs `totalRecords` to render. But for aggregated queries (`group_by`) we can't know the total number of groups without fetching every bucket, and we ruled out approximations.

So this PR adds a `cursor` pagination mode that doesn't need a total, just the current page and how many pages ahead are known to exist.

(Internal) Shortcut: [sc-61255](https://app.shortcut.com/opendatasoft/story/61255).

## What changes

`Pagination` becomes a discriminated union on `kind`:

- `kind: 'numbered'` (or omitted): the existing pager, unchanged.
- `kind: 'cursor'`: a new pager `‹ … [n-1] [n] [n+1] … ›`, no total, no first/last jump.

Cursor mode takes a single integer, `pagesAhead` (how many pages after the current one are known to exist). The component derives the rest: `>= 1` shows the next page number and enables `›`, `>= 2` shows the trailing ellipsis. The left side needs nothing (we navigated through those pages, so they exist). Hosts compute `pagesAhead` with a sentinel fetch (ask for a few extra rows and divide), documented in `types.ts`.

Two small supports: `Numbering.svelte` drops the `/total` part when there's no total, and `Pagination` takes an optional `displayedRecords` so the "X-Y" range ends on the real last row of a partial page (`41-43`, not `41-50`).

**No breaking changes.** `kind` defaults to numbered, all current usages keep working.

**Changelog:** Table: new `kind: 'cursor'` pagination mode to paginate without a known total record count.

## To test

In Storybook, `Table/Pagination`:

1. `Cursor (no total)`: page 1 shows `‹ [1] 2 … ›` (`‹` disabled), range `1-5` with no total. Navigate to the last page: trailing ellipsis drops one page early, `›` disables, range shows the real row count.
2. `Cursor mid-page`: starts page 3, both ellipses visible.
3. Numbered stories (`Paginated`, `PageSize`, `CustomStyle`) still behave as before.


RafaelSzmarowski (approved):
LGTM ! Works well and code is very readable ! Good job

PR #299, opened 2026-06-18
feat(Map): Add onMapReady callback to transmit the map instance

richterb:
## Summary

The goal for this PR is to expose the Map instance to calling code, to open for more customization.

### Changes
- Add an optional `onMapReady` option, which can be used by the calling code to get the `Map` instance when it is ready. It can then be used to add event hooks, use directly some Maplibre methods, etc.

There is already a lot of Maplibre exposed through the `WebGlMap` component, such as the layers/source configuration, so I think it makes sense, and leaves the component as a "supercharged Maplibre".

It is used in a platform PR mentioned in the feed below.

## Review checklist

- [ ] Description is complete


wmai (approved):
Clean and minimal addition. LGTM 👏

PR #300, opened 2026-06-22
fix(Table): Clean overflow with tooltip when header labels are too long

richterb:
## Summary

The goal for this PR is to handle headers overflow in situations where the header cells are limited in width.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-62014](https://app.shortcut.com/opendatasoft/story/62014).

### Changes
The issue doesn't appear in "pure" SDK use, because the header cells can take as much width as they need. However, in the platform, a `max-width` appears and can lead to broken overflow.

In this PR:
- We make sure the label itself can't overflow, and show an ellipsis; the sort icon button, if present, is still visible.
- I leverage `tooltipOnOverflow` to only show a tooltip when the header title is too long. I raised it one level higher and make sure it can work without a specific cell format. This felt like the most consistent, code and design-wise.

#### Breaking Changes
None in theory. The HTML structure changes a bit, but accessibility/semantic-wise it shouldn't have any impact.

## To be tested
- There is a new story dedicated to long titles, it should work as intended
- For regular labels, nothing changes
- For labels too long in the platform, it should cut clean, display an ellipsis, and a tooltip with the entire value

## Review checklist

- [ ] Description is complete


packages/visualizations/src/components/Table/Headers/SortButton.svelte
```diff
@@ -32,6 +32,10 @@
         justify-content: space-between;
     }
 
+    :global(button svg) {
+        flex-shrink: 0;
+    }
```
cybervinvin: One minor comment: `:global(button svg)` targets all `button svg` elements on the page, not just those inside this component. The style itself is harmless, but a scoped selector (e.g. `.sort-button button svg`) would be safer. Non-blocking, worth a follow-up.

cybervinvin (approved):
Tested the five cases manually (ellipsis on sortable and plain headers, tooltip on hover, sort icon visibility at narrow widths, sort interaction). All working as expected.

packages/visualizations/src/components/Table/Headers/SortButton.svelte
richterb: Indeed the `:global` kills the scoping and this ended up compiled as `button svg` which applied everywhere on the page, which is bad 😅 

The `global` is there initially so that Svelte doesn't automatically prune it because it can't see a `svg` within the same component. 

I changed to `button :global(svg)` so that the Svelte scoping applies automatically on the button while still leaving the `svg` alone, this ends up as:
Screenshot 2026-07-09 at 10 50 39

PR #301, opened 2026-07-08
feat: support top corner positioning for the POI map legend

wmai:
## Summary

The goal for this PR is to let the POI map display its category legend as a floating overlay in the top corner of the map, instead of always below it. The Explorer map (platform side) opens as a full-bleed map, and a legend pushed below the map does not work in that layout. A corner legend sits on the map itself and works at any size.

(Internal for Huwise only) Associated Shortcut ticket: [sc-61879](https://app.shortcut.com/huwise/story/61879).

### Changes

- `CategoryLegend` gets an optional `position` value with one new option: `top-left`. With it, the POI map renders the legend as a floating box over the map (white card, shadow, scroll when too tall). Any other value keeps the current rendering below the map, so existing consumers see no change.
- The overlay uses logical CSS insets, so it mirrors automatically in RTL (`top-left` becomes visually top-right), as the ticket asks. There is no `top-right` option: the map controls already live in that corner, and the RTL mirroring is a behaviour, not a second option.
- The legend stays visible in fullscreen: `WebGlMapOptions` gets an optional `fullscreenContainer` element. The POI map passes its own wrapper (map + overlay) by default, and a host that draws its own overlays around the map passes the element holding them instead. MapLibre only reads the container when the control is built, so the control is rebuilt when the element arrives.
- A Storybook story shows the corner legend.

### Taking styling back from the consumers

The platform was overriding MapLibre internals to get results the map can express itself, out-specifying our own rules to win. Three fixes, so a host does not have to reach inside the map to size or frame it:

- **`aspectRatio: 'auto'` on the POI map**, the same contract the chart already has. The map then fills the height its parent gives it instead of deriving it from its width. `Card` already carried the `ods-dataviz--fill` modifier, the POI map simply never asked for it, and it re-applied `DEFAULT_ASPECT_RATIO` when the option was omitted, so a consumer had no way to turn the ratio off through the API.
- **Map corners are rounded by the map**, via `border-radius: var(--map-border-radius, 6px)` on the map container. MapLibre already clips that element, so the canvas follows. A host drawing the map edge to edge sets the variable to `0` instead of overriding `.maplibregl-map`. Deliberately not wired to `--visualization-card-border-radius`: the canvas sits inside the card padding, it is not the card corner, and that variable is themable.
- **The control group gets its `margin-left`**, next to the `margin-top` and `margin-right` it already had, so the group keeps the same distance from the edge whatever corner it sits in.
- **The map controls are styled here now**, in a new `styles/map-controls.css` imported next to `variables.css`. The POI map, the choropleth map and a bare WebGl map each mount their own MapLibre instance, so a component-scoped rule would have to be written three times, and it was indeed written in three places, one of them a consumer's stylesheet. Radius, padding, button size and hover apply to maps framed as a visualization card; a bare WebGl map keeps MapLibre's own sizing, as before.
- **The choropleth back button and its navigation strip carry their own look.** `BackButton` had `position: absolute` and nothing else, so consumers had to give it a size, a background and a position, the last one through a formula encoding our control geometry. It now looks like the control buttons it sits under, `MapRender` places it (only the map knows which corner its controls are in), and the offset reads `--map-control-button-size` so it cannot drift. Same for the navigation strip separator and the mini-map buttons.
- **`controlPosition` on `WebGlMapOptions`** so a host that overlays something in a corner can move the controls to the opposite one instead of shifting MapLibre's container in CSS. MapLibre reads the position when a control is added, so mounted controls are taken off and put back; controls the map deliberately hides, such as while a modal popup covers that corner, stay hidden.

### Fix: a card border could not be turned off, only made invisible

A host drawing a visualization edge to edge would set `--visualization-card-border: transparent` and still lose a pixel on each side, because that only changes the colour and the border keeps its place in the layout. Width is now its own variable, `--visualization-card-border-width`, the way the KPI card already separates `--kpi-card-border-color` from `--kpi-card-border-width`. It defaults to `1px`, so nothing moves unless a host asks for `0`.

### Fix: the links menu opened outside the view in right-to-left

Reported on an embed, and it predates this branch. `LinksMenu` anchored its panel with `right: 0`, a physical property, while the trigger is placed by a flex header and therefore follows reading order. In right-to-left the trigger is on the physical left, so a panel anchored to the physical right stayed pinned to the right edge of a 32px box and spilled 168px to the left, out of the visualization and under the edge of the iframe.

The panel now uses `inset-inline-end: 0`: it lines its reading-end edge up with the trigger and grows toward the reading-start, so inward, in both directions. The vertical offset stays physical, it does not mirror.

`KpiCard` had to move with it. It is the only host that places the trigger itself, and it did so with `right: 1rem`, which in right-to-left left the button on the reading-**start** side, where a panel growing toward the start would have spilled the other way. It is now `inset-inline-end: 1rem`, putting the button on the reading-end side like the card header does, which is also where a menu affordance belongs.

Two mirroring slips fixed in passing: the action menu items were `text-align: left` where an anchor item aligns from the reading side on its own, and the choropleth mini-map strip spaced its buttons with `margin-left`.

A `RightToLeft` story renders a chart and a KPI card inside `dir="rtl"`, since the two place the trigger differently.

#### Breaking Changes

None in the API: every new option is optional and the defaults reproduce the current behaviour. `aspectRatio` only widens its type, a number behaves as before.

**Consumers overriding map chrome in their own stylesheets should have a look.** The styles now shipped here are the ones the Huwise platform was applying, to the pixel (same spacing scale, same radius, same `#fff` hover, same 24px buttons), so its own rules could simply be deleted. A consumer that was applying *different* values with a selector weaker than these will now see ours instead. The values sit in `styles/map-controls.css` and read from the variables in `styles/variables.css`.

**Two things move visibly, in right-to-left reading order only**, and both are the fix rather than a side effect: the links menu panel opens inward instead of outward, and the KPI card menu button sits on the reading-end side of the card instead of the physical right.

### Changelog

_POI map: the category legend can now be displayed as a floating overlay in the top corner of the map (`position: 'top-left'`), including in fullscreen. POI map: `aspectRatio` accepts `'auto'` to fill the parent height, like the chart. WebGL map: new optional `fullscreenContainer` option to expand a custom element in fullscreen, new optional `controlPosition` to choose the corner the controls sit in, and map corners rounded through the `--map-border-radius` variable. Maps now ship the chrome of their own controls, of the choropleth back button and of the choropleth navigation strip, instead of leaving it to consumers._

## To be tested

- **Nothing should have moved**, and that is the main thing: existing POI map stories keep their legend below the map, their fullscreen and their rounded corners, and the control chrome on a POI map, a choropleth map and a bare `WebGlMap` renders exactly as today (only the stylesheet it comes from changed). Same for the choropleth back button and navigation strip.
- **The corner legend**: the `PoiMapLegendTopLeft` story, in fullscreen too, and mirrored in RTL (`dir="rtl"`).
- **The new options**: `aspectRatio: 'auto'` inside a fixed-height parent, `controlPosition` moving both controls together (including with a modal popup open), `fullscreenContainer` pointing at an element wider than the map, and `--map-border-radius: 0` / `--visualization-card-border-width: 0` on an ancestor.
- **RTL**: the `LinksMenu / RightToLeft` story, both menus opening inward, and any existing left-to-right story unchanged.

## Review checklist

- [ ] Description is complete
- [ ] Commits respect the Conventional Commits Specification

### DIFF
--- a/packages/visualizations/src/components/Legend/types.ts
+++ b/packages/visualizations/src/components/Legend/types.ts
@@ -70,21 +70,16 @@ export type CategoryItem =
 
 export const CATEGORY_LEGEND_POSITION = {
     /** Below the map (default, legacy behaviour). */
-    Bottom: 'bottom',
-    /** Floating overlay in the map's top inline-start corner. */
-    TopLeft: 'top-left',
-    /** Floating overlay in the map's top inline-end corner. */
-    TopRight: 'top-right',
+    bottom: 'bottom',
+    /** Floating overlay in the map's top inline-start corner (mirrors in RTL). */
+    topLeft: 'top-left',
 } as const;
 
 // Includes the legacy chart legend positions (bottom/left/right) so the shared
 // CategoryLegend type stays compatible with the chart's custom legend. The map
-// only acts on 'top-left' / 'top-right' (overlay); other values fall back to the
-// default below-the-map rendering.
-export type CategoryLegendPosition =
-    | LegendPositions
-    | typeof CATEGORY_LEGEND_POSITION.TopLeft
-    | typeof CATEGORY_LEGEND_POSITION.TopRight;
+// only acts on 'top-left' (overlay); other values fall back to the default
+// below-the-map rendering.
+export type CategoryLegendPosition = LegendPositions | typeof CATEGORY_LEGEND_POSITION.topLeft;
 
 export type CategoryLegend = {
     type: 'category';
@@ -93,10 +88,9 @@ export type CategoryLegend = {
     align?: 'start' | 'center' | 'end';
     /**
      * Where the legend renders relative to the map. Defaults to `'bottom'`
-     * (below the map). `'top-left'` / `'top-right'` render it as a floating
-     * overlay in the map's top corner. Positioning is logical (inline-start /
-     * inline-end), so `'top-left'` automatically mirrors to the top-right corner
-     * in RTL layouts (and vice versa).
+     * (below the map). `'top-left'` renders it as a floating overlay in the
+     * map's top corner. Positioning is logical (inline-start), so it
+     * automatically mirrors to the top-right corner in RTL layouts.
      */
     position?: CategoryLegendPosition;
 };
--- a/packages/visualizations/src/components/Legend/types.ts
+++ b/packages/visualizations/src/components/Legend/types.ts
@@ -68,9 +68,35 @@ export type CategoryItem =
     | LineCategoryItem
     | ImageCategoryItem;
 
+export const CATEGORY_LEGEND_POSITION = {
+    /** Below the map (default, legacy behaviour). */
+    Bottom: 'bottom',
+    /** Floating overlay in the map's top inline-start corner. */
+    TopLeft: 'top-left',
+    /** Floating overlay in the map's top inline-end corner. */
+    TopRight: 'top-right',
+} as const;
+
+// Includes the legacy chart legend positions (bottom/left/right) so the shared
+// CategoryLegend type stays compatible with the chart's custom legend. The map
+// only acts on 'top-left' / 'top-right' (overlay); other values fall back to the
+// default below-the-map rendering.
+export type CategoryLegendPosition =
+    | LegendPositions
+    | typeof CATEGORY_LEGEND_POSITION.TopLeft
+    | typeof CATEGORY_LEGEND_POSITION.TopRight;
+
 export type CategoryLegend = {
     type: 'category';
     items: CategoryItem[];
     title?: string;
     align?: 'start' | 'center' | 'end';
+    /**
+     * Where the legend renders relative to the map. Defaults to `'bottom'`
+     * (below the map). `'top-left'` / `'top-right'` render it as a floating
+     * overlay in the map's top corner. Positioning is logical (inline-start /
+     * inline-end), so `'top-left'` automatically mirrors to the top-right corner
+     * in RTL layouts (and vice versa).
+     */
+    position?: CategoryLegendPosition;
 };
--- a/packages/visualizations/src/components/Map/Poi/Poi.svelte
+++ b/packages/visualizations/src/components/Map/Poi/Poi.svelte
@@ -36,8 +36,9 @@
 
     $: cssVarStyles = `--aspect-ratio:${aspectRatio};`;
 
-    // The fullscreen control expands this wrapper (map + legend overlay) rather
-    // than the bare map container, so the overlay legend stays visible.
+    // Default fullscreen target: this wrapper holds the map and the legend overlay, so both expand
+    // together. A host rendering its own overlays around the map passes the element holding them as
+    // `fullscreenContainer`, and that one takes over.
     let mainEl: HTMLElement | undefined;
 </script>
 
@@ -58,7 +59,10 @@
         aria-describedby={description ? mapId.toString() : undefined}
     >
         {#key options.style}
-            <WelGlMap options={{ ...options, fullscreenContainer: mainEl }} data={data.value} />
+            <WelGlMap
+                options={{ ...options, fullscreenContainer: options.fullscreenContainer ?? mainEl }}
+                data={data.value}
+            />
         {/key}
         {#if legend && isOverlayLegend}
             <div class="legend-overlay">
--- a/packages/visualizations/src/components/Map/Poi/Poi.svelte
+++ b/packages/visualizations/src/components/Map/Poi/Poi.svelte
@@ -29,6 +29,11 @@
     // default) keeps the legacy legend below the map.
     $: isOverlayLegend = !!legend && legend.position === CATEGORY_LEGEND_POSITION.topLeft;
 
+    // 'auto' takes the height the parent gives us instead of deriving it from the
+    // width, for a host that sizes the map itself. `aspect-ratio: auto` is what CSS
+    // already calls "no imposed ratio", so the same variable carries both cases.
+    $: isAutoAspectRatio = aspectRatio === 'auto';
+
     $: cssVarStyles = `--aspect-ratio:${aspectRatio};`;
 
     // The fullscreen control expands this wrapper (map + legend overlay) rather
@@ -44,9 +49,11 @@
     style={cssVarStyles}
     tag="figure"
     className="map-card maps-container ods-dataviz--maps"
+    fill={isAutoAspectRatio}
 >
     <div
         class="main"
+        class:fill={isAutoAspectRatio}
         bind:this={mainEl}
         aria-describedby={description ? mapId.toString() : undefined}
     >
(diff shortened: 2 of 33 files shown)


packages/visualizations/src/components/Legend/types.ts
```diff
@@ -68,9 +68,35 @@ export type CategoryItem =
     | LineCategoryItem
     | ImageCategoryItem;
 
+export const CATEGORY_LEGEND_POSITION = {
+    /** Below the map (default, legacy behaviour). */
+    Bottom: 'bottom',
```
RafaelSzmarowski: Nitpicking but I think we usually do lowercase here ? 
```suggestion
    bottom: 'bottom',
```

packages/visualizations/src/components/Legend/types.ts
```diff
@@ -68,9 +68,35 @@ export type CategoryItem =
...
+    /** Floating overlay in the map's top inline-start corner. */
+    TopLeft: 'top-left',
+    /** Floating overlay in the map's top inline-end corner. */
+    TopRight: 'top-right',
```
RafaelSzmarowski: I guess you did this to handle RTL but I have a serious doubt if we want to do it as the Maplibre seems always displayed on the top right... Until we can mirror them on the left is it a real improvement to be able to have the legend also on the top right ? Should we just and always display it on top left maybe ?

packages/visualizations/src/components/Map/Poi/Poi.svelte
```diff
@@ -56,6 +74,27 @@
         position: relative;
         display: block;
     }
+    /* Floating legend overlaid on the map. Positioned with logical insets so it
+       mirrors automatically in RTL (top-left <-> top-right). The map controls sit
+       in the top-inline-end corner, so the legend keeps clear of them. */
```
RafaelSzmarowski: MapLibre’s FullscreenControl expands the map canvas by default, so the overlay legend (a sibling inside .main) is left out of fullscreen and therefore not visible. 

I don't know if it's worth the effort but if we want to fix it we can probably add a `fullscreenContainer` option to WebGl / MapPOI so the control can expand a custom element, then in Poi.svelte bind a ref to .main and pass it so fullscreen includes the map and the overlay legend. Also set aspect-ratio: auto on .main:fullscreen (and .maplibregl-pseudo-fullscreen fallback used by MapLibre when the Fullscreen API is unavailable (iOS)) so the wrapper can fill the screen. Rebuild the control in a setFullscreenContainer function in WebGl.ts if the container is set after it’s already on the map, since MapLibre only reads container at construction.

RafaelSzmarowski (commented):
It works quiet well ! Nice job ! 

I just wonder if we need to offer a top-right position option and also if we want to fix the fullscreen mode. Please feel free to give your opinion on this 😃

packages/visualizations/src/components/Legend/types.ts
wmai: Good catch, aligned with `LEGEND_POSITIONS` above. (7bc99d8)

packages/visualizations/src/components/Legend/types.ts
wmai: Agreed, dropped. The ticket asks for one position (top-left in LTR, top-right in RTL), so RTL is handled by the logical insets rather than a second option. The RTL overlap with the controls stays, better fixed by making the control position configurable, which the design wants anyway (bottom-right). (7bc99d8)

packages/visualizations/src/components/Map/Poi/Poi.svelte
wmai: Worth it, and your sketch was spot on, thanks: new `fullscreenContainer` option, `Poi.svelte` passes its wrapper, control rebuilt when the element arrives (iOS fallback covered). The legend stays visible in fullscreen now. (7bc99d8)

wmai:
Thanks @RafaelSzmarowski for the review! All three are in (7bc99d8): lowercase keys, `top-right` dropped, and fullscreen fixed with your `fullscreenContainer` idea. Description updated. Only the RTL controls overlap is left, as a follow-up.

wmai:
A lot a changes in this PR. It started as a legend floating in the corner of the map (still here), but getting the platform to draw a map edge to edge showed it was reaching into our own map markup to do it: the zoom buttons, the choropleth back button, the corner rounding, forcing a height. None of that should be a consumer's job, so it moved here.

A host can now just ask for what it needs: `aspectRatio: 'auto'` to fill the parent height like the chart, `controlPosition` for the corner the buttons sit in, `fullscreenContainer` so full screen expands an element of your choice, `--map-border-radius`, and `--visualization-card-border-width` so a card border can be turned off instead of only made invisible.

Two right-to-left bugs fixed, both older than this branch: the links menu opened outside the visualization, and the KPI card menu button sat on the wrong side. There's a `RightToLeft` story for both.

packages/visualizations/src/components/utils/Card.svelte
RafaelSzmarowski, line 48: Not directly linked to your changes, but as this PR seems to mix fixes it might be worth changing here. Seems like `box-sizing: border-box` defined in `.card.ods-dataviz—fill` could make more sense here in `.card` and fix the bad display in our POI stories where the right border is hidden. It's an optional fix, so if we are unsure we can leave as it this and not block this PR.

Image

packages/visualizations/src/components/utils/BackButton.svelte
```diff
@@ -10,12 +10,26 @@
...
         align-items: center;
+        box-sizing: content-box;
+        width: var(--map-control-button-size);
+        height: var(--map-control-button-size);
+        padding: var(--spacing-50);
+        border: none;
+        border-radius: var(--border-radius-6);
+        background-color: var(--background-color);
+        box-shadow: 0 1px 6px 0 rgba(0, 0, 0, 0.26);
     }
```
RafaelSzmarowski, line 34: Also not direclty linked to the PR but we could probably flip the button in RTL ? 

```
button:dir(rtl) {
    transform: scaleX(-1);
}
```

RafaelSzmarowski (approved):
LGTM ! Indeed the PR evolved to mix different fixes, all make sense in my opinion. I've left two more things I've spotted while reviewing, if we want to move faster and fix them later it's good to me as well.

packages/visualizations/src/components/utils/Card.svelte
wmai: Done. I checked on the platform, nothing changes there, it already applies `box-sizing` everywhere. It only matters in our stories. (60dfe20)

packages/visualizations/src/components/utils/BackButton.svelte
wmai: Done, exactly as you wrote it. Chromatic never sees it, the button only shows after you click a region. I checked left to right and right to left. (cb2b056)

PR #302, opened 2026-07-17
feat(Table): add field type icons in column headers

cybervinvin:
<sub>Disclaimer: Code & description drafted by AI, human-assisted for extra flavor. ;)</sub>

## Summary

The goal for this PR is to display a field type icon next to each column header label in the Table component, giving users an at-a-glance visual cue about the data type in each column.

(Internal for Opendatasoft only) Associated Shortcut ticket: [sc-59790](https://app.shortcut.com/opendatasoft/story/59790).

### Changes

- **`FieldTypeIcon.svelte`** — new component rendering an inline SVG icon for each `DataFormat` value. Icons sourced from the Figma Perceptio icon set. Short-text, long-text, and unknown formats fall back to the "abc" icon.
- **`Th.svelte`** — header title wrapped in `span.th-title-content` (`inline-flex`, gap) hosting the icon and label side by side.
- **`showRowNumbers` prop** — opt-in boolean (`default: false`) threaded from `TableCard` → `Table` → `Headers` + `Body` → `Row`, adding a muted sequential row-number column on the left (including loading skeleton).
- **5 new `DATA_FORMAT` entries** — `json`, `file`, `image`, `ipAddress`, `id` — with `FormatPropsTypeMap` mappings and `Column` union types. `file`/`image` → `URLFormat`; `json`/`ip-address`/`id` → `TextFormat`.
- **Stories** — `FieldTypeIcons`, `NoRowNumbers`, and `AllColumnTypes` (all 12 column types on one row) replace the old `LongColumnTitles` story.

#### Breaking Changes

No breaking change — `showRowNumbers` defaults to `false`, preserving the current layout for existing tables.

### Changelog

In the Table component: column headers now display a field type icon. A new `showRowNumbers` option (opt-in, defaults to `false`) adds a sequential row number column on the left. Five new column types are available: `json`, `file`, `image`, `ipAddress`, `id`.

## Open discussion

- **New types not in `TRUNCATING_FORMATS`**: `json`, `file`, `image`, `ip-address`, and `id` get the `text-overflow: ellipsis` CSS treatment but won't show a tooltip on overflow. Intentional (values are typically short), or should they be added?

## To be tested

- All 12 column types display their correct icon in the header.
- Tables with `showRowNumbers: true` show a sequential row-number column, including in loading skeleton rows.
- Tables without `showRowNumbers` (or with `showRowNumbers: false`) render identically to the previous behaviour.
- Sticky columns and RTL direction not broken by the new header structure.

## Review checklist

- [ ] Description is complete


packages/visualizations/src/components/Table/Headers/Th.svelte
```diff
@@ -88,4 +95,10 @@
     :global(.ods-dataviz--default .hover.sticky) {
         border-left: none;
     }
+
+    .th-title-content {
+        display: inline-flex;
+        align-items: center;
+        gap: var(--spacing-50);
+    }
```
wmai, line 110: I tested the `LongColumnTitles` story and found a problem: long titles are not cut with "..." anymore. The full title goes over the next column title, so both become unreadable (see screenshot). Before this PR, the title was cut and a tooltip showed the full text on hover.

packages/visualizations/src/components/Table/Row.svelte
```diff
@@ -35,6 +37,9 @@
     on:focusin={rowProps && handleMouseEnter}
     on:focusout={rowProps && handleMouseLeave}
 >
+    {#if showRowNumbers}
+        <td class="row-number-cell" aria-hidden="true">{rowIndex + 1}</td>
```
wmai: I noticed a problem when row numbers are used together with pinned (sticky) columns. When you scroll to the right, the pinned columns slide over the row numbers, and the numbers disappear. This happens because the row number column is not part of the pinned system, while the button column next to it is.

I think the row numbers should stay visible when scrolling, since that is when they are the most useful. If you prefer to fix that later, that works too, but then we should write the limitation somewhere and add a story that shows this case.

wmai: Small product question: with pagination, the numbers start again at 1 on every page. So "row 3" points to a different record on each page. Is that what we want, or should page 2 continue with 11, 12, 13...?

packages/visualizations/src/components/Table/Headers/Headers.svelte
```diff
@@ -5,10 +5,14 @@
     export let columns: Column[];
     export let extraButtonColumn = false;
     export let extraButtonColumnLabel: string | undefined;
+    export let showRowNumbers = false;
 </script>
 
 <thead>
     <tr>
+        {#if showRowNumbers}
+            <th class="row-number-header" aria-hidden="true" />
```
wmai: The row numbers are hidden from screen readers (here and on the cells in `Row.svelte:41`). So a user with a screen reader will not know the numbers exist. It feels strange because these users are probably the ones who benefit the most from knowing their position in the table.

packages/visualizations/src/components/Table/Cell/CellContent.svelte
```diff
@@ -63,6 +63,20 @@
                 {...getOptions(column, record)}
                 debugWarnings={$debugWarnings}
             />
+        {:else if isColumnOfType(column, DATA_FORMAT.file) || isColumnOfType(column, DATA_FORMAT.image)}
```
wmai: This condition does exactly the same thing as the `url` just above. They could be merged into one condition (`url`, `file` or `image`).

packages/visualizations/src/components/Table/Body.svelte
```diff
@@ -51,4 +57,16 @@
     .button-cell__empty {
         min-width: 28px;
     }
+
+    :global(.ods-dataviz--default .row-number-cell) {
+        text-align: right;
```
wmai: I am not 100% sure about this one, but in right-to-left mode the number may end up glued to the edge of the table, without space, because the cell mixes a "right" alignment with a direction-aware padding. Worth a quick look at the RTL story with row numbers turned on. If confirmed, the fix should be tiny (`text-align: end`).

packages/visualizations/src/components/Table/Cell/CellContent.svelte
```diff
@@ -63,6 +63,20 @@
...
+            <URLFormat
+                value={getValue(column, record)}
+                {...getOptions(column, record)}
+                debugWarnings={$debugWarnings}
+            />
+        {:else if isColumnOfType(column, DATA_FORMAT.json)
```
wmai: One thing to keep in mind for the `json` type: if the value is a real object (which is what the API returns for json fields), the cell will show "[object Object]". The story does not catch it because it passes the value already as a string.

So either we convert the value to a string inside the component, or we say clearly that the caller must do it. It depends on how the platform integration is planned, so this can also be a follow-up.

wmai (changes requested):
Nice work ! The icons look great and match the platform ones perfectly. I left a few small suggestions, plus one real bug with long column titles.

packages/visualizations/src/components/Table/Headers/Th.svelte
cybervinvin: Good catch. The `th-title` span lost its `overflow: hidden` and `text-overflow: ellipsis` when I restructured it into a flex container for the icon. Restored.

packages/visualizations/src/components/Table/Row.svelte
cybervinvin: Fixed. The row-number column now has `position: sticky; inset-inline-start: 0`. Added a `--row-number-width` CSS custom property on the scrollbox to offset all other sticky columns correctly. The border and shadow follow the same visual logic as the existing sticky system: border always visible when row-number is the only sticky column, shadow appears on scroll.

cybervinvin: Good question. Went with continuous numbering: page 2 starts at 11 (with 10 records/page). Added a `rowOffset` prop computed from `(current - 1) * recordsPerPage` in `TableCard`, passed down to `Row`.

packages/visualizations/src/components/Table/Headers/Headers.svelte
cybervinvin: Fixed. Removed `aria-hidden` from both the header cell and the body cells. The header keeps its `aria-label="Row number"`.

packages/visualizations/src/components/Table/Cell/CellContent.svelte
cybervinvin: Fixed. Merged into a single `url || file || image` condition.

packages/visualizations/src/components/Table/Body.svelte
cybervinvin: Confirmed and fixed. Changed `text-align: right` to `text-align: end`.

packages/visualizations/src/components/Table/Cell/CellContent.svelte
cybervinvin: Fixed inside the component with `JSON.stringify` when the value is an object. Updated the story to pass a real object (`{ active: true, score: 98 }`) so the serialization is actually tested.

packages/visualizations/src/components/Table/Headers/Headers.svelte
```diff
@@ -5,10 +5,14 @@
     export let columns: Column[];
     export let extraButtonColumn = false;
     export let extraButtonColumnLabel: string | undefined;
+    export let showRowNumbers = false;
 </script>
 
 <thead>
     <tr>
+        {#if showRowNumbers}
+            <th class="row-number-header" scope="col" aria-label="Row number" />
```
RafaelSzmarowski: aria-label is hardcoded in english, should probably be a localized label

packages/visualizations/src/components/Table/Headers/Th.svelte
```diff
@@ -44,10 +45,16 @@
                 on:click={column.onClick}
                 labels={column?.sortLabels}
             >
-                <span class="th-title" use:tooltipOnOverflow>{column.title}</span>
+                <span class="th-title-content">
+                    <FieldTypeIcon dataFormat={column.dataFormat} />
+                    <span class="th-title" use:tooltipOnOverflow>{column.title}</span>
+                </span>
```
RafaelSzmarowski, line 54: `FieldTypeIcon` is always rendered, it's a breaking change. I have the feeling that it would feel more natural to make this optional.

packages/visualizations-react/stories/Table/PaginatedTemplates.tsx
```diff
@@ -63,6 +66,7 @@ export const PaginatedTemplate = (pagination: Pagination) => {
 
     const stateFulOptions = {
         ...options,
+        showRowNumbers,
```
RafaelSzmarowski, line 70: I don't understand the goal of this change as this does not show the number on rows in my storybook.

This on the contrary shows them: 
 
```suggestion
        showRowNumbers: true,
```

Am I missing some purpose ?

packages/visualizations/src/components/Table/Headers/FieldTypeIcon.svelte
```diff
@@ -0,0 +1,172 @@
...
+        />`,
+        // id-custom (Figma)
+        [DATA_FORMAT.id]: `<path d="M7 6H9V18H7V6Z"
+        /><path fill-rule="evenodd" clip-rule="evenodd"
+            d="M11 6H13C16.3137 6 19 8.68629 19 12C19 15.3137 16.3137 18 13 18H11V6ZM13 8V16C15.2091 16
+            17 14.2091 17 12C17 9.79086 15.2091 8 13 8Z"
+        />`,
+    };
+
+    $: icon = (dataFormat && paths[dataFormat]) ?? TEXT;
```
RafaelSzmarowski: Just wonder if a "?" icon would be better than the Text one ?

packages/visualizations/src/components/Table/actions.ts
```diff
@@ -3,7 +3,12 @@ import type { Instance } from 'tippy.js';
 import { DATA_FORMAT } from './constants';
 
 // These are the three column types whose CSS truncates content (see CellContent.svelte styles)
```
RafaelSzmarowski:
```suggestion
// These are the four column types whose CSS truncates content (see CellContent.svelte styles)
```

RafaelSzmarowski (changes requested):
Few remarks but everything works really fine, the most important thing that we can discuss is whether we really want to force FieldTypeIcon to be always displayed or make it an option for retro compatibility and more user options ?

packages/visualizations/src/components/Table/Headers/Headers.svelte
cybervinvin: Fixed. Added `rowNumberLabel?: string` to `TableOptions` (defaults to `'Row number'`), passed down to `Headers.svelte`. Consumers can now pass a localized label.

packages/visualizations/src/components/Table/Headers/Th.svelte
cybervinvin: Fixed. Added `showFieldTypeIcons?: boolean` to `TableOptions` (opt-in, defaults to `false`). The icon is only rendered when the prop is explicitly set to `true`. Updated the `FieldTypeIcons` story accordingly.

packages/visualizations-react/stories/Table/PaginatedTemplates.tsx
cybervinvin: `PaginatedTemplate` is shared across multiple stories: some want row numbers (`RtlDirection`, `FieldTypeIcons`), others don't (`Numbered`, `Cursor`). The dynamic prop is intentional so each story controls it via Storybook args. Also added `showFieldTypeIcons` as a parallel prop for consistency.

packages/visualizations/src/components/Table/Headers/FieldTypeIcon.svelte
cybervinvin: Good call. Used the question-mark icon from Figma as the fallback for unknown/undefined types.

packages/visualizations/src/components/Table/actions.ts
cybervinvin: Fixed.

packages/visualizations/src/components/Table/Headers/Th.svelte
```diff
@@ -44,10 +46,18 @@
...
+                    {#if showFieldTypeIcons}
+                        <FieldTypeIcon dataFormat={column.dataFormat} />
+                    {/if}
+                    <span class="th-title" use:tooltipOnOverflow>{column.title}</span>
+                </span>
             </SortButton>
         {:else}
-            <div class="th-title" use:tooltipOnOverflow>{column.title}</div>
+            <span class="th-title-content">
+                <FieldTypeIcon dataFormat={column.dataFormat} />
```
RafaelSzmarowski: You display conditionally the icon in the sortable columns but the non-sortable branch still always renders the icon here

packages/visualizations-react/stories/Table/Table.stories.tsx
```diff
@@ -147,9 +149,126 @@ const longTitleData: Async<TableData> = {
...
+            file: 'https://example.com/report.pdf',
+            img: 'https://example.com/photo.jpg',
+            ip: '192.168.1.42',
+            id: 'usr_abc123',
+        },
+    ],
+    loading: false,
+};
+
+export const AllColumnTypes: StoryObj<typeof Table> = {
```
RafaelSzmarowski, line 256: This story do not pass `showFieldTypeIcons: true` it displays the icons because this are non sortable icons but when you fix the point above you'll have to pass the prop here.

packages/visualizations/src/components/Table/actions.ts
```diff
@@ -2,8 +2,13 @@ import tippyAction from 'components/utils/tippy';
 import type { Instance } from 'tippy.js';
 import { DATA_FORMAT } from './constants';
 
-// These are the three column types whose CSS truncates content (see CellContent.svelte styles)
-const TRUNCATING_FORMATS: string[] = [DATA_FORMAT.shortText, DATA_FORMAT.longText, DATA_FORMAT.url];
+// These are the four column types whose CSS truncates content (see CellContent.svelte styles)
+const TRUNCATING_FORMATS: string[] = [
```
RafaelSzmarowski, line 6: If my understanding is good, the CSS truncate all those column types: 

```
:global(.ods-dataviz--default div.table-data--long-text > span),
    :global(.ods-dataviz--default div.table-data--short-text),
    :global(.ods-dataviz--default div.table-data--json),
    :global(.ods-dataviz--default div.table-data--ip-address),
    :global(.ods-dataviz--default div.table-data--id),
    :global(.ods-dataviz--default div.table-data--url),
    :global(.ods-dataviz--default div.table-data--file),
    :global(.ods-dataviz--default div.table-data--image)
```

But here the list TRUNCATING_FORMATS decides which cells get an overflow tooltip, this means both are desync so missing types here are truncated but have no tooltip which seems wrong


```suggestion
// Column types whose CSS truncates content (see CellContent.svelte styles)
const TRUNCATING_FORMATS: string[] = [
    DATA_FORMAT.shortText,
    DATA_FORMAT.longText,
    DATA_FORMAT.url,
    DATA_FORMAT.json,
    DATA_FORMAT.file,
    DATA_FORMAT.image,
    DATA_FORMAT.ipAddress,
    DATA_FORMAT.id,
];
```

packages/visualizations-react/stories/Table/Table.stories.tsx
```diff
@@ -147,9 +149,126 @@ const longTitleData: Async<TableData> = {
...
+        },
+    },
+    render: (args: TableProps) => (
+        <div style={{ maxWidth: '900px' }}>
+            <Table {...args} />
+        </div>
+    ),
+};
+
+const allTypesColumns: Column[] = [
```
RafaelSzmarowski, line 168: We can also add the unknown type as well now

packages/visualizations/src/components/Format/types.ts
```diff
@@ -35,6 +35,12 @@ export type TextFormatProps = {
     debugWarnings?: boolean;
 };
 
+export type JsonFormatProps = {
+    value: string | Record<string, unknown>;
+    valueToLabel?: (v: string | Record<string, unknown>) => string;
+    debugWarnings?: boolean;
+};
+
```
RafaelSzmarowski, line 48: The type here seems innacurate as in CellContent we apply JSON.stringify

```suggestion
/**
 * JSON column format props.
 * Objects are `JSON.stringify`'d in CellContent before display; `valueToLabel`
 * therefore always receives the displayed string, never the raw object.
 */
export type JsonFormatProps = {
    value: string | Record<string, unknown>;
    valueToLabel?: (v: string) => string;
    debugWarnings?: boolean;
};
```

packages/visualizations/src/components/Table/Headers/Headers.svelte
```diff
@@ -5,18 +5,31 @@
...
         {#each columns as column (column.key)}
-            <Th {column} />
+            <Th {column} {showFieldTypeIcons} />
         {/each}
     </tr>
 </thead>
 
 <style>
+    .row-number-header {
+        width: 2rem;
```
RafaelSzmarowski: Here and in the hardcoded --row-number-width (2rem) The fixed 2rem width do not handle properly large numbers with sticky column rows, I'm not sure how much we want to fix it, in our use case our API always only fetch max 10 000 I think :

Image

We have few choices 

1. The most heavy probably a bit overkill: make the width to adapt: 
- Headers measures the row-number <th> with ResizeObserver (avoids bind:clientWidth sticky issues)
- Width is bound up to Table as --row-number-width
- Header uses min-width: 2rem so it can grow with wide numbers

2. Probbably the best easy win here: increase the width to 3 rem for example
3. Ellipsis + tooltip

RafaelSzmarowski (changes requested):
Thanks for the changes ! I saw few new points that needs our attention before merging ! Please have a look and feel free to discuss them.

packages/visualizations/src/components/Table/Headers/Th.svelte
cybervinvin: Fixed. `.th-title-content` now uses flex with `min-width: 0`, restoring the ellipsis truncation. Verified in the `Long Column Titles` story.

packages/visualizations/src/components/Table/Row.svelte
cybervinvin: Fixed. The row-number column is now part of the sticky system (`position: sticky`), so it stays visible when scrolling with pinned columns. Added the `StickyColumnsWithRowNumbers` story to demo it.

cybervinvin: Fixed. Row numbers now use a `rowOffset` derived from the current page, so they continue across pages instead of resetting. Added the `NumberedWithRowNumbers` story.

packages/visualizations/src/components/Table/Headers/Headers.svelte
cybervinvin: Fixed. The row-number cells are no longer hidden from assistive tech, they render as plain text content.

packages/visualizations/src/components/Table/Cell/CellContent.svelte
cybervinvin: Merged into one condition in `CellContent.svelte`, thanks.

packages/visualizations/src/components/Table/Body.svelte
cybervinvin: Fixed with `text-align: end`, plus direction-aware border/shadow CSS for the row-number column.

packages/visualizations/src/components/Table/Cell/CellContent.svelte
cybervinvin: Fixed. Objects are now `JSON.stringify`'d before display.

packages/visualizations/src/components/Table/Headers/Th.svelte
cybervinvin: Fixed. Wrapped in `{#if showFieldTypeIcons}` to match the sortable branch.

packages/visualizations-react/stories/Table/Table.stories.tsx
cybervinvin: Added `showFieldTypeIcons: true` to the story options.

packages/visualizations/src/components/Table/actions.ts
cybervinvin: Applied your suggestion, added `file`/`image`/`ip-address`/`id`.

packages/visualizations-react/stories/Table/Table.stories.tsx
cybervinvin: Added an `Unknown` column to `AllColumnTypes`. That also surfaced that `CellContent` had no fallback branch for an unrecognized `dataFormat` (rendered an empty cell), so I added one.

packages/visualizations/src/components/Format/types.ts
cybervinvin: Applied your suggestion.

packages/visualizations/src/components/Table/Headers/Headers.svelte
cybervinvin: Went with option 2 (3rem) given the ~10k row cap you mentioned. Testing it also surfaced two related issues: the border/shadow was always visible instead of only on scroll, and once restricted to the last sticky column, a 1px gap between multiple pinned columns let scrolled content show through. Both fixed.

packages/visualizations/src/components/Table/Row.svelte
```diff
@@ -35,6 +38,9 @@
     on:focusin={rowProps && handleMouseEnter}
     on:focusout={rowProps && handleMouseLeave}
 >
+    {#if showRowNumbers}
+        <td class="row-number-cell">{rowOffset + rowIndex + 1}</td>
+    {/if}
```
RafaelSzmarowski, line 43: The row-number cell is emitted before the action (zoom in platform) cell, and the same order is hardcoded in Headers.svelte and in the loading rows of Body.svelte. 


Image

Our design has the action button on the left of the row number, and there is currently no way to get that order from the options.

Image

Could we swap the two blocks in those three places? If the order is expected to differ per host, an option would work too.

packages/visualizations/src/components/Table/Table.svelte
```diff
@@ -56,19 +63,31 @@
     }
 </script>
 
-<div class="scrollbox" bind:this={scrollBox} on:scroll={handleScroll}>
+<div
+    class="scrollbox"
+    bind:this={scrollBox}
+    on:scroll={handleScroll}
+    style:--row-number-width={showRowNumbers ? 'calc(3rem + var(--spacing-75))' : '0px'}
```
RafaelSzmarowski, line 70: We already discussed this earlier and I thought a 3rem would be sufficient but while testing on the platform table I noticed that this won't be sufficient: 

`--row-number-width` is hardcoded to calc(3rem + var(--spacing-75)) which resolves to 57px, but it is what every sticky column uses as its left offset in _sticky.scss. This breaks as soon as the host constrains column widths, because nothing guarantees the row-number column actually renders at that width.

Concretely, in the platform the table stylesheet applies min-width: 75px to all th/td. .row-number-header only declares width: 3rem, and min-width wins over width, so the column really renders at 75px while the sticky columns still latch at 57px. The action column then overlaps the last 18px of the row-number column, exactly where the digits sit, since .row-number-cell is text-align: end with a trailing padding. Because td has an opaque white background and both cells share z-index: 10, the later one in the DOM paints on top and the numbers become invisible. It only shows up once the table is scrolled horizontally, since unscrolled sticky cells sit at their natural position.

Image

Two ways out, whichever you prefer: measure the rendered width instead of assuming it (the sticky store already measures column widths through stickyColumnsWidth.updateColumn) it seems the more adapted fix but the heaviest, or make the row-number column width non-overridable by a host, e.g. a min-width in a :global rule on .row-number-header so it cannot drift from the hardcoded value, but this will limit the possibilites for host and our platform design, so I personally would more lean toward the first one.

Note this is not visible in the current stories, which do not constrain column widths but we should maybe take this into account in the SDK stories as well.

RafaelSzmarowski (changes requested):
I noticed last few points while testing those changes inside the platform: the number and action cells are not in same order as our design and the fixed width shows some limits when hosts override the column widths. I'll let you have a look on this, best way is to test directly in platform as this is the real goal behind the changes.

packages/visualizations/src/components/Table/Row.svelte
cybervinvin: Fixed. Row-number now registers as a regular sticky column ordered after the action button (Row.svelte, Headers.svelte, and the loading skeleton), instead of a hardcoded always-first position.

packages/visualizations/src/components/Table/Table.svelte
cybervinvin: Went with your first option: row-number's width is now measured from the rendered element via ResizeObserver, kept in sync through host CSS changes and window resize, instead of the hardcoded `--row-number-width`. That CSS variable is gone; sticky columns get their offset directly from measured widths.

Also found and fixed the same `min-width` issue on the action column while testing in platform, plus a couple of related visual bugs (gap masking, header background).

packages/visualizations/src/components/Table/Body.svelte
```diff
@@ -27,6 +29,9 @@
                 {#if rowProps?.onClick}
                     <td class="button-cell__empty" />
                 {/if}
+                {#if showRowNumbers}
+                    <td class="row-number-cell row-number-cell--loading" />
```
RafaelSzmarowski: I noticed two loading state issues when data is slow to arrive. Both come from the same place: loading rows build raw <td> instead of going through <Td>, so they skip the measured sticky offset and the loaded cells box model.

- The loading row number cell is sticky at inset-inline-start: 0, while the header and loaded rows use the measured offset after the action column. On horizontal scroll they overlap.
- Those same raw cells don’t get the action-column padding, the .cell-content padding, or column.sticky on LoadingCell. The skeleton is therefore a different width/height than the loaded table.

To fix this, we should probably render loading rows through the same <Td> path, including column on LoadingCell.svelte. Also give `.button-cell__empty` the same 28×28 size as the action button, and drop the inset-inline-start: 0 workaround.

```
{#each Array(loadingRowsNumber) as _}
            <tr>
                {#if rowProps?.onClick}
                    <!-- Action column has no loading indicator, but keeps the button's footprint -->
                    <Td>
                        <div class="button-cell__empty" />
                    </Td>
                {/if}
                {#if showRowNumbers}
                    <Td rowNumber />
                {/if}
                {#each columns as column}
                    <LoadingCell {column} />
                {/each}
            </tr>
        {/each}
```

Because currently we have this on scroll and the table resizes after data is loaded: 

Image

RafaelSzmarowski (changes requested):
Thanks for the latest changes, I have a last remark that is a nice to have and will avoid possible loading state transition glitches

packages/visualizations/src/components/Table/Body.svelte
cybervinvin: Fixed. Loading rows now render through `Td` (and `LoadingCell` takes a `column` prop), same path as loaded rows. The action and row-number placeholders get the measured sticky offset, `.button-cell__empty` matches the button's 28x28 footprint, and a sticky data column stays sticky during loading too.

RafaelSzmarowski (approved):
LGTM ! Thanks for the last change !

Just we need to fix the Lint errors but otherwise all works well !

wmai (approved):
Nicely done ! LGTM 🚀

PR #322, opened 2026-09-10
feat: Table URL detection + Pagination groupPageControls option

cybervinvin:
## Summary

This PR extends URL auto-linkification to `short-text`/`long-text` table columns, matching the existing behavior already available on `url` columns, and adds a `groupPageControls` option to the Pagination component to group its page and page-size controls flush right.

🏴‍☠️ Quick fix for the playground's record table — no Shortcut ticket.

### Changes

**Table: URL detection**

`CellContent.svelte` now routes `short-text`/`long-text` columns through `URLFormat` instead of `TextFormat`. A column can opt out via the new `disableUrlDetection` option (added to `TextFormatProps`), falling back to `TextFormat` as before.

`URLFormat` gained a `warnOnInvalidUrl` prop (default `true`). `CellContent` passes `warnOnInvalidUrl={false}` for text columns, so a non-URL string never logs a warning — only genuine `url` columns still warn on invalid values (unchanged).

`isValidUrl` (in `utils.ts`) rejects strings longer than `MAX_URL_LENGTH` (2048 chars, exported) before calling `new URL()`, since that constructor's cost grows with input size.

Tests added in `Table.test.tsx`: linkification and plain-text rendering for both column types, the `disableUrlDetection` opt-out, unchanged `url`-column behavior, debug-warning suppression, and length-guard boundary at exactly 2048/2049 chars.

**Pagination: groupPageControls**

Adds `groupPageControls` to `NumberedPagination`: when `true`, the `pages` and `page-size` grid areas size to content (`auto`) and sit flush right, while `numbering` takes the remaining width (`1fr`) — no change to area order or content, only sizing/positioning. Default (`false`) keeps the existing 3-equal-column layout unchanged. Below the 500px container breakpoint, the existing 2-row stacked layout applies regardless of `groupPageControls`.

Tests added in `Table.test.tsx` (class presence for both states), plus a `GroupedControls` Storybook story under `Table/Pagination` (required threading the option through `PageSizeTemplate`, which previously only forwarded an explicit field list).

#### Breaking Changes

None for the public API on either feature — both are new optional additions.

Behaviorally, for URL detection: any existing `short-text`/`long-text` column whose values look like URLs will start rendering as clickable links, since detection is opt-out rather than opt-in. Set `disableUrlDetection: true` on a column to keep the previous plain-text rendering.

No behavioral change for Pagination: `groupPageControls` defaults to `false`.

### Changelog

In the Table component: short-text and long-text columns now detect and linkify URL values automatically, with a `disableUrlDetection` column option to opt out. The Pagination component also gains a `groupPageControls` option to group its page and page-size controls flush right.

## Open discussion

- Routing text columns through `URLFormat` also silences the pre-existing per-cell "not a valid text" debug warning (`debugWarnings`) on `short-text`/`long-text` — confirmed as an intended side effect, not accidental.
- `MAX_URL_LENGTH` (2048) is a practical cutoff (legacy IE convention), not a spec limit. Confirmed there's no known need for longer URLs in text columns today; the constant can be raised later without new API surface if that changes.
- URL detection defaults to opt-out (`disableUrlDetection: true` to disable) rather than opt-in — confirmed as the intended behavior, so it activates automatically on every existing `short-text`/`long-text` column on upgrade.
- The narrow-container behavior (<=500px) for `groupPageControls` was left at the implementer's discretion per the request; the choice made was to keep the existing stacked layout unchanged rather than adapt it further. Worth confirming this is sufficient.

## To be tested

Any Studio/platform table already using `short-text`/`long-text` columns with URL-like content should be spot-checked after this ships, since those cells will start rendering as links.

## Review checklist

- [x] Description is complete


packages/visualizations/src/components/Format/utils.ts
```diff
@@ -20,7 +20,14 @@ export function warn(value: unknown, format: string, debugWarnings = false) {
     }
 }
 
+// Legacy (IE-era) URL length convention; skip new URL() (cost grows with input size) past this size.
+export const MAX_URL_LENGTH = 2048;
+
 export function isValidUrl(text: unknown): text is string {
+    if (typeof text === 'string' && text.length > MAX_URL_LENGTH) {
```
wmai, line 27: I have a small worry about the 2048 limit. This PR now checks every text cell: "is this a URL?". Text can be very long, so it makes sense to stop trying after a certain size. The limit is inside `isValidUrl`, but that function is also used for columns that are already links (`url`, `file`, `image`). So a long file link would stop being a link too, even though we already know it is a URL.

For text columns, I'd still keep a limit, maybe just a much higher number so we only skip the huge values. For url columns, we might not need a check at all. If you'd rather keep one there too, that same higher number would still let a real long URL through.

wmai (changes requested):
Nicely done ! The tests look good and I also tried a few nasty values (`javascript:`, `data:`, `blob:`) and none of them became a link 👍

I just have one small worry about the 2048 length limit.

packages/visualizations/src/components/Format/utils.ts
cybervinvin: Good catch, thanks for digging into this.

Fixed by decoupling the two: `isValidUrl` now takes an optional `maxLength` param instead of applying the constant unconditionally. `url`/`file`/`image` columns pass nothing, so they keep their original unbounded behavior. `short-text`/`long-text` still pass `MAX_URL_LENGTH` (2048), since that's the case where most values aren't URLs at all and the perf guard actually matters.

Added a regression test covering `url`/`file`/`image` with a value over 2048 chars to lock this in.

wmai (approved):
Thanks for the quick fix, LGTM 👏

I published `0.37.1-beta.0` on npm. You can pin it on #15439 and switch `useColumns.ts` to `short-text` from there.

PR #323, opened 2026-09-23
fix(Table): define tooltip width and height for images before loading

RafaelSzmarowski:
## Summary

The goal for this PR is to fix the tooltip appearance before the image starts loading. Previously the tooltip had no width and height before the image loads so it could seemed broken. I've reused same background color as we already have for skeletons.

(Internal for Huwise only) Associated Shortcut ticket: [sc-63117](https://app.shortcut.com/huwise/story/63117/bug-sdk-table-image-tooltip-has-no-loading-state-and-looks-broken-while-a-large-image-loads).

### Changes

Add a fixed height and width for the empty tooltip container.

Before: 

https://github.com/user-attachments/assets/198cb053-bfb5-4c07-b201-c68e8e80c1a7

After:

https://github.com/user-attachments/assets/34da340f-05d0-4013-ab73-295dcb53bd8d


RafaelSzmarowski:
Made a last change to avoid introducing bad styling in small images: removed the grey background to avoid it creating a grey border around small images and fixed the height and width to avoid resizing:

https://github.com/user-attachments/assets/627d138d-fd21-45eb-9305-3b556c7b430e

